PR #19, opened 2016-07-20
implement more indexing methods for `World`

adRichter:
add type PillarIndex
add chunk_at
add chunk_from_pillar

### DIFF
--- a/base/src/world/mod.rs
+++ b/base/src/world/mod.rs
@@ -35,10 +35,15 @@ pub const CHUNK_SIZE: u16 = 16;
 /// PillarIndexComponent::min_value())`² many hex pillars.
 pub type PillarIndexComponent = math::AxialType;
 
-/// A new-type to index chunks. This is different from the `AxialPoint` type
-/// which always represents a pillar position. So two different `AxialPoint`s
+/// A new-type to index Pillars. Always represents a specific Pillar in
+/// absolute world coordinates.
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
+pub struct PillarIndex(pub math::AxialPoint);
+
+/// A new-type to index chunks. This is different from the 'PillarIndex' type
+/// which always represents a pillar position. So two different 'PillarIndices'
 /// could refer to two pillars in the same chunk, while two different
-/// `ChunkIndex`es always refer to two different chunks.
+/// 'ChunkIndices' always refer to two different chunks.
 #[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
 pub struct ChunkIndex(pub math::AxialPoint);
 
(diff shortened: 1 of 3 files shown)


base/src/world/mod.rs
```diff
@@ -35,10 +35,15 @@ pub const CHUNK_SIZE: u16 = 16;
...
+/// A new-type to index Pillars. Always represents a specific Pillar in
+/// absolute world coordinates.
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
+pub struct PillarIndex(pub math::AxialPoint);
+
+/// A new-type to index chunks. This is different from the 'PillarIndex' type
+/// which always represents a pillar position. So two different 'PillarIndices'
 /// could refer to two pillars in the same chunk, while two different
-/// `ChunkIndex`es always refer to two different chunks.
+/// 'ChunkIndices' always refer to two different chunks.
```
LukasKalbertodt: This was intentional, actually. Rust documentation is written in Markdown in which inline code blocks are marked with back-ticks. Those are `[Shift] + [the key right to the '?' key]` (at least on German keyboards). You used the single quotes `'` instead, which don't have a special meaning in markdown. 

I'd say you just revert this line change, ok?

base/src/world/provider.rs
```diff
@@ -34,3 +34,28 @@ impl Provider for NullProvider {
         false
     }
 }
+
+/// A fallback provider that holds two chunk providers with one beeing primary
```
LukasKalbertodt: Typo :bee:

base/src/world/provider.rs
```diff
@@ -34,3 +34,28 @@ impl Provider for NullProvider {
...
+
+impl<P: Provider, F: Provider> Provider for FallbackProvider<P, F> {
+    fn load_chunk(&self, pos: ChunkIndex) -> Option<Chunk> {
+        if self.primary.is_chunk_loadable(pos) {
+            self.primary.load_chunk(pos)
+        } else if self.fallback.is_chunk_loadable(pos) {
+            self.fallback.load_chunk(pos)
+        } else {
+            NullProvider {}.load_chunk(pos)
+        }
```
LukasKalbertodt: This can be simplified: you don't need the `NullProvider` here. Just return what the fallback provider returned if the primary can't. If the fallback can't provide either, it will return `None`, so no need to ask the `NullProvider`.

base/src/world/world.rs
```diff
@@ -25,9 +25,9 @@ impl World {
         World { chunks: chunks }
     }
 
-    /// Returns the hex pillar at the given world position, iff the
+    /// Returns the hex pillar at the given world position, if the
```
LukasKalbertodt: And this was intentional, too :P 

"iff" is an abbreviation for "if and only if" :wink:

PR #22, opened 2016-07-21
Implement `ChunkProvider` for `WorldGenerator`

adRichter:


base/src/gen/world/mod.rs
```diff
@@ -1,6 +1,8 @@
 //! Procedurally generating the game world.
 //!
 
+use super::super::world::{Chunk, ChunkIndex, Provider};
```
jonas-schievink: `use world::...` should work

base/src/gen/world/mod.rs
```diff
@@ -18,3 +20,17 @@ impl WorldGenerator {
         self.seed
     }
 }
+
+impl Provider for WorldGenerator {
+    /// Returns requested `Chunk`.
```
jonas-schievink: Documentation on trait impls isn't really necessary unless there's something special about the specific implementation - The trait's docs will be displayed if none are present.

client/src/game.rs
```diff
@@ -1,4 +1,4 @@
-use base::world::Provider as WorldProvider;
+use base::world::ChunkProvider as WorldProvider;
```
jonas-schievink: The `as WorldProvider` should be removed eventually, but you don't need to do this now

PR #23, opened 2016-07-21
startup config change

NiklasPaulRothe:


client/Cargo.toml
```diff
@@ -8,6 +8,9 @@ version = "0.1.0"
 [dependencies]
 glium = "0.15.0"
 log = "0.3.6"
+clap = "*"
```
jonas-schievink: `*` versions should not be used, set it to the most up-to-date version of `clap` instead

jonas-schievink: You can also run `cargo add clap` while inside the client crate to do it automatically

client/src/config.rs
```diff
@@ -16,6 +35,32 @@ impl Default for Config {
...
+
+        .arg(Arg::with_name("Resolution")
+            .help("(e.g. =1280x720) 'Sets Resolution to new value'")
+            .takes_value(true)
+            .long("resolution")
+        )
+        .get_matches();
+
+    if let Some(ref res) = matches.value_of("Resolution"){
+        let reg_res = Regex::new(r"^[1-9]\d{1,4}x[1-9]\d{1,4}").unwrap();
```
jonas-schievink: You don't need to use a regex for this, the parsing logic will work just fine if you handle errors properly

client/src/config.rs
```diff
@@ -16,6 +35,32 @@ impl Default for Config {
...
+        if reg_res.is_match(res){
+            let mut split_res = res.split("x");
+            let vec_res = split_res.collect::<Vec<&str>>();
+            let res_x = vec_res[0].parse::<u32>().unwrap();
+            let res_y = vec_res[1].parse::<u32>().unwrap();
+            toml_config.resolution = Dimension2::new(res_x, res_y);
+        }else{}
```
jonas-schievink: This is badly formatted and probably fails Travis. Also, as said above, not necessary.

client/src/config.rs
```diff
@@ -16,6 +35,32 @@ impl Default for Config {
...
+            .long("resolution")
+        )
+        .get_matches();
+
+    if let Some(ref res) = matches.value_of("Resolution"){
+        let reg_res = Regex::new(r"^[1-9]\d{1,4}x[1-9]\d{1,4}").unwrap();
+        if reg_res.is_match(res){
+            let mut split_res = res.split("x");
+            let vec_res = split_res.collect::<Vec<&str>>();
+            let res_x = vec_res[0].parse::<u32>().unwrap();
```
jonas-schievink: You shouldn't `unwrap` this, but provide a useful error message to the user

client/src/config.rs
```diff
@@ -16,6 +35,32 @@ impl Default for Config {
...
+            let res_y = vec_res[1].parse::<u32>().unwrap();
```
jonas-schievink: `vec_res[1]` might not exist if there's no `x` in there, so this can panic

client/Cargo.toml
LukasKalbertodt: The current version of [`clap`](https://crates.io/crates/clap) is 2.9.2, so write `clap = "2" instead.

client/src/config.rs
jonas-schievink: (well, this is guaranteed to succeed if you use the regex, not sure if that's a good idea though)

client/src/config.rs
jonas-schievink: (here's where a good error message should be returned, if you decide to still use the regex)

plantex/main.rs
```diff
@@ -18,7 +18,15 @@ fn main() {
 
 
     info!("~~~~~~~~~~ Plantex started ~~~~~~~~~~");
-    let res = client::start_game(client::Config::load_config(), &base::world::NullProvider);
+    let conf: ::client::Config;
```
jonas-schievink: `let conf = match ...`

and in the match

`Ok(v) => v,`

jonas-schievink:
You have a few unused imports that fail Travis

jonas-schievink:
Okay, we should use the seed of the config instead of the hardcoded 42 we use now. Also needs a rebase (worldgen integration landed).

client/src/config.rs
```diff
@@ -1,21 +1,106 @@
...
+            .takes_value(true)
+            .long("seed"))
+        .get_matches();
+
+    if let Some(res) = matches.value_of("Resolution") {
+        let reg_res = Regex::new(r"^([1-9]\d{1,4})x([1-9]\d{1,4})").unwrap();
+
+        if reg_res.is_match(res) {
+            for cap in reg_res.captures_iter(res) {
+                let res_x = cap.at(1).unwrap().parse::<u32>().unwrap();
```
jonas-schievink: Ah `at(x).unwrap()` is actually the same as direct indexing. Just using `cap[0]` should be fine now, since you're using the regex to parse it.

Also the indices look off-by-one to me (unless `at` is 1-based)?

client/src/config.rs
```diff
@@ -21,27 +21,12 @@ pub struct Config {
                     * Chunkweite */
 }
 impl Config {
+    ///creates a new Config in three steps:
```
jonas-schievink: Put a space after the `///`

client/src/config.rs
```diff
@@ -60,11 +45,54 @@ fn command_config(mut toml_config: Config) -> Result<Config, Box<StdError>> {
...
+        let conf_final = match config_command(t_conf, &matches){
+            Ok(n) => n,
+            Err(e) => return Err(e),
+        };
+        Ok(conf_final)
+    }
+}
+
+impl Default for Config {
+    ///Returns default config from hard coded values
```
jonas-schievink: No need to document impls or methods of well-defined traits

client/src/config.rs
NiklasPaulRothe: does not work the same, actually panicked

at ist 1-based

client/src/config.rs
```diff
@@ -60,11 +45,54 @@ fn command_config(mut toml_config: Config) -> Result<Config, Box<StdError>> {
...
+            resolution: Dimension2::new(800, 600),
+            window_mode: WindowMode::Windowed,
+            window_title: format!("Plantex {}", env!("CARGO_PKG_VERSION")),
+            vsync: false,
+            seed: 42,
+        }
+    }
+}
+
+fn config_toml(mut default_config: Config, matches: &ArgMatches<>) -> Result<Config, Box<StdError>>{
```
jonas-schievink: Drop the `<>` after `ArgMatches`

jonas-schievink:
This needs a few warnings fixed

client/src/config.rs
```diff
@@ -1,21 +1,159 @@
...
+use std::fs::File;
+use std::io::Read;
+use self::toml::Parser;
 
 pub struct Config {
     pub resolution: Dimension2u,
     pub window_mode: WindowMode,
     pub window_title: String,
+    pub vsync: bool,
+    pub seed: u64, /* sichtweite
```
jonas-schievink: English comments please :)

client/src/config.rs
```diff
@@ -1,21 +1,159 @@
...
+                .long("seed"))
+            .arg(Arg::with_name("File")
+                .help("Takes config file")
+                .takes_value(true)
+                .long("config-file"))
+            .get_matches();
+
+        let conf = Config::default();
+
+        let t_conf = match config_toml(conf, &matches) {
```
jonas-schievink: Does `let t_conf = try!(config_toml(conf, &matches));` work?

client/src/config.rs
```diff
@@ -1,21 +1,159 @@
...
+            Ok(n) => n,
+            Err(e) => return Err(e),
+        };
+        let conf_final = match config_command(t_conf, &matches) {
```
jonas-schievink: `try!` using `try!` here, too

client/src/config.rs
```diff
@@ -1,21 +1,256 @@
...
     pub resolution: Dimension2u,
     pub window_mode: WindowMode,
     pub window_title: String,
+    pub vsync: bool,
+    pub seed: u64, /* view range
+                    * anti aliasing
+                    * Controls
+                    * frames
+                    * seed
+                    * Chunkweite */
```
jonas-schievink: still german

client/src/config.rs
```diff
@@ -1,21 +1,256 @@
...
+            "#;
+
+            let mut f = match File::create("config.toml") {
+                Ok(n) => n,
+                _ => return Err("Failed writing config file (e.g. missing write permission".into()),
+            };
+            match f.write_all(&toml.to_string().into_bytes()) {
+                Ok(n) => n,
+                _ => return Err("Failed writing config file (e.g. missing write permission".into()),
+            };
```
LukasKalbertodt: You can use `try!` here. Just write `let mut f = try!(File::create("..."));`. No need to create your own error string, as the `io::Error` returned by `File::create` implements `StdError`.

client/src/config.rs
```diff
@@ -1,21 +1,256 @@
...
+            vsync: false,
+            seed: 42,
         }
     }
 }
 
+fn config_toml(mut default_config: Config, matches: &ArgMatches) -> Result<Config, Box<StdError>> {
+    let mut name = "config.toml";
+    if let Some(file) = matches.value_of("File") {
+        let file_reg = Regex::new(".*\\.toml").unwrap();
```
LukasKalbertodt: Prefer raw strings: `r#".*\.toml"#`

PR #26, opened 2016-07-21
Implement World Rendering

### LINKED ISSUE #18: Basic world rendering
The world should be rendered as hex pillars and not just ugly triangle (like it's done now). Whoever tries to solve this issue needs to dig into some dirty code (sorry, I hadn't had enough time). 

You need to create vertex and index buffers and do some OpenGL Stuff. You can also clean up the code.

NilsAffing:
Closes #18 
Implemented the rendering of all Hexagons per Chunk.
ToDo:
- Pillartransformation (scalevalue, translation)

### DIFF
--- a/base/src/world/mod.rs
+++ b/base/src/world/mod.rs
@@ -28,7 +28,7 @@ pub const PILLAR_STEP_HEIGHT: f32 = 0.5;
 
 /// How many hex pillars a chunk is long. So the number of hex pillars in a
 /// chunk is `CHUNK_SIZE`².
-pub const CHUNK_SIZE: u16 = 16;
+pub const CHUNK_SIZE: u16 = 32;
 
 /// This type is used to index into one dimension of the world. Thus we can
 /// "only" index `(PillarIndexComponent::max_value() -
--- a/base/src/world/mod.rs
+++ b/base/src/world/mod.rs
@@ -28,7 +28,7 @@ pub const PILLAR_STEP_HEIGHT: f32 = 0.5;
 
 /// How many hex pillars a chunk is long. So the number of hex pillars in a
 /// chunk is `CHUNK_SIZE`².
-pub const CHUNK_SIZE: u16 = 32;
+pub const CHUNK_SIZE: u16 = 16;
 
 /// This type is used to index into one dimension of the world. Thus we can
 /// "only" index `(PillarIndexComponent::max_value() -
--- a/base/src/world/mod.rs
+++ b/base/src/world/mod.rs
@@ -15,16 +15,16 @@ pub use self::hex_pillar::*;
 pub use self::provider::*;
 pub use self::world::World;
 
-/// Inner radius of the hexagons
-pub const HEX_INNER_RADIUS: f32 = 1.5;
-
 /// Outer radius of the hexagons (from center to corner)
-pub const HEX_OUTER_RADIUS: f32 = HEX_INNER_RADIUS * (::math::SQRT_3 / 2.0);
+pub const HEX_OUTER_RADIUS: f32 = 5.0;
+
+/// Inner radius of the hexagons
+pub const HEX_INNER_RADIUS: f32 = HEX_OUTER_RADIUS * (::math::SQRT_3 / 2.0);
 
 /// The height of the hex pillars is discretized. So instead of saving a `f32`
 /// to represent the height, we have fixed steps of heights and we will just
 /// save a `u16` to represent the height.
-pub const PILLAR_STEP_HEIGHT: f32 = 0.5;
+pub const PILLAR_STEP_HEIGHT: f32 = 5.0;
 
 /// How many hex pillars a chunk is long. So the number of hex pillars in a
 /// chunk is `CHUNK_SIZE`².
--- a/base/src/world/mod.rs
+++ b/base/src/world/mod.rs
@@ -16,7 +16,7 @@ pub use self::provider::*;
 pub use self::world::World;
 
 /// Outer radius of the hexagons (from center to corner)
-pub const HEX_OUTER_RADIUS: f32 = 5.0;
+pub const HEX_OUTER_RADIUS: f32 = 1.0;
 
 /// Inner radius of the hexagons
 pub const HEX_INNER_RADIUS: f32 = HEX_OUTER_RADIUS * (::math::SQRT_3 / 2.0);
@@ -24,7 +24,7 @@ pub const HEX_INNER_RADIUS: f32 = HEX_OUTER_RADIUS * (::math::SQRT_3 / 2.0);
 /// The height of the hex pillars is discretized. So instead of saving a `f32`
 /// to represent the height, we have fixed steps of heights and we will just
 /// save a `u16` to represent the height.
-pub const PILLAR_STEP_HEIGHT: f32 = 5.0;
+pub const PILLAR_STEP_HEIGHT: f32 = 0.5;
 
 /// How many hex pillars a chunk is long. So the number of hex pillars in a
 /// chunk is `CHUNK_SIZE`².
--- a/client/src/world/chunk.rs
+++ b/client/src/world/chunk.rs
@@ -46,10 +46,7 @@ impl ChunkView {
         }
 
         // convert to vector
-        let mut final_buffer = Vec::new();
-        for element in hexagon_vertex_buffer {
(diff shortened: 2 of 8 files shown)


base/src/world/mod.rs
```diff
@@ -15,16 +15,16 @@ pub use self::hex_pillar::*;
...
+pub const HEX_OUTER_RADIUS: f32 = 5.0;
+
+/// Inner radius of the hexagons
+pub const HEX_INNER_RADIUS: f32 = HEX_OUTER_RADIUS * (::math::SQRT_3 / 2.0);
 
 /// The height of the hex pillars is discretized. So instead of saving a `f32`
 /// to represent the height, we have fixed steps of heights and we will just
 /// save a `u16` to represent the height.
-pub const PILLAR_STEP_HEIGHT: f32 = 0.5;
+pub const PILLAR_STEP_HEIGHT: f32 = 5.0;
```
jonas-schievink: This means that we'd only be able to deal in 5-meter steps of pillars

client/src/world/chunk.rs
```diff
@@ -17,19 +28,32 @@ impl ChunkView {
...
+
+        }
+
+
+        let mut final_buffer = Vec::new();
+        for element in hexagon_vertex_buffer {
+            final_buffer.push(element);
+        }
+
+        // println!("{:#?}", final_buffer);
```
jonas-schievink: Hier könnt ihr `debug!` verwenden (oder löschen)

client/src/world/chunk.rs
```diff
@@ -17,19 +27,32 @@ impl ChunkView {
...
-                          Vertex {
-                              position: [0.0, 1.0, 0.0],
-                              color: [0.0, 0.0, 1.0],
-                          }];
-        let vbuf = glium::VertexBuffer::new(facade, &buffer).unwrap();
+
+        // Create one hexagon for this chunk
+        let mut hexagon_vertex_buffer = VecDeque::new();
+        for i in 0..6 {
+            let raw_modell = hex_corner(world::HEX_OUTER_RADIUS, i);
```
LukasKalbertodt: You can destructure the tuple here like this:
```rust
let (x, y) = hex_corner(...);
```

This avoids the ugly `.0` and so on

client/src/world/chunk.rs
```diff
@@ -17,19 +27,32 @@ impl ChunkView {
...
+
+            hexagon_vertex_buffer.push_front(Vertex {
+                position: [raw_modell.0, raw_modell.1, world::PILLAR_STEP_HEIGHT / 2.0],
```
LukasKalbertodt: This should just be `PILLAR_STEP_HEIGHT` I'd say...

client/src/world/chunk.rs
```diff
@@ -17,19 +27,32 @@ impl ChunkView {
...
+                color: [0.15 * i as f32, 0.0, 0.0],
+            });
+
+            hexagon_vertex_buffer.push_back(Vertex {
+                position: [raw_modell.0, raw_modell.1, -1.0 * (world::PILLAR_STEP_HEIGHT / 2.0)],
```
LukasKalbertodt: One of the two vertices that are added should have `z = 0`. We want a vertex buffer layout that stands on the x-y-plane.

client/src/world/chunk.rs
```diff
@@ -17,19 +27,32 @@ impl ChunkView {
...
+                color: [1.0 - 0.15 * i as f32, 0.15 * i as f32, 0.0],
+            });
+
+        }
+
+        // convert to vector
+        let mut final_buffer = Vec::new();
+        for element in hexagon_vertex_buffer {
+            final_buffer.push(element);
+        }
```
LukasKalbertodt: All of this can be replaced by

```rust
let final_buffer: Vec<_> = hexagon_vertex_buffer.iter().collect();
```

client/src/world/chunk.rs
```diff
@@ -44,40 +67,65 @@ impl ChunkView {
             }
         }
 
+        // Indecies
+        let raw_index_buffer = [5, 0, 1, 2, 5, 1, 4, 5, 2, 3, 4, 2 /* TOP */, 6, 7, 8, 8, 9,
+                                6, 9, 11, 6, 9, 10, 11 /* BOTTOM */, 6, 5, 4, 7, 6, 4, 6, 0,
+                                5, 11, 0, 6, 10, 1, 0, 11, 10, 0, 9, 2, 1, 10, 9, 1, 8, 3, 2, 9,
+                                8, 2, 7, 4, 3, 8, 7, 3 /* Body */];
```
LukasKalbertodt: Have you tried to add `#[rustfmt_skip]` above this line? Then `rustfmt` shouldn't mess with your custom formatting.

jonas-schievink: That only works with `#[feature(custom_attributes)]`, right?

client/src/world/chunk.rs
```diff
@@ -44,40 +67,65 @@ impl ChunkView {
...
 
-        let ibuf = glium::index::NoIndices(glium::index::PrimitiveType::TrianglesList);
-
         for pillar_pos in &self.pillars_positions {
             let uniforms = uniform!{
+              scale_matrix:[[1.0,0.0,0.0,0.0]
+                      ,[0.0,1.0,0.0,0.0],
+                      //FIXME 5.0 heightvalue
+                      [0.0,0.0,1.0,0.0],
+                      [0.0,0.0,0.0,1.0f32]],
```
LukasKalbertodt: You can use
```rust
scale_matrix: Matrix4::from_scale(1.0).to_arr()
```
Or rather the height value than `1.0`.

client/src/world/chunk.rs
jonas-schievink: And `#[feature(stmt_macros)]` or whathever it's called

client/src/world/chunk_std.frag
```diff
@@ -4,8 +4,8 @@ in vec3 x_color;
 out vec4 color;
 
 uniform vec2 offset;
+uniform mat4 scale_matrix;
```
LukasKalbertodt: It's not necessary to add it here if you don't use it in the fragment shader

client/src/world/pillar.rs
```diff
@@ -1,5 +1,5 @@
+use base::world;
 
 /// Graphical representation of the `base::HexPillar`.
 pub struct PillarView {
-
 }
```
LukasKalbertodt: This file is useless, right? If so, please remove with `git rm path/to/file.rs`

client/src/world/chunk.rs
LukasKalbertodt: Ah fuck, you're right. Ok, so ignore all those comments :P

NilsAffing: Does not work

client/src/world/chunk.rs
NilsAffing: Gives an error

client/src/world/chunk.rs
NilsAffing: We want to scale Only the Z(height) Value later.

LukasKalbertodt:
And please add a short description to your PR (I already added the "closes") and adjust the title if necessary.

client/src/world/chunk.rs
jonas-schievink: Why not replace the matrix with a single height value? That way you don't waste so much uniform storage for a single value, and the shaders get simpler, too.

client/src/world/chunk.rs
```diff
@@ -90,38 +92,40 @@ impl ChunkView {
     pub fn draw<S>(&self, surface: &mut S, camera: &Camera)
         where S: glium::Surface
     {
+        unsafe {
```
jonas-schievink: :bomb:

base/src/world/mod.rs
```diff
@@ -28,7 +28,7 @@ pub const PILLAR_STEP_HEIGHT: f32 = 0.5;
 
 /// How many hex pillars a chunk is long. So the number of hex pillars in a
 /// chunk is `CHUNK_SIZE`².
-pub const CHUNK_SIZE: u16 = 16;
+pub const CHUNK_SIZE: u16 = 32;
```
jonas-schievink: Why this?

PR #27, opened 2016-07-21
Camera and Event Handler

FlorianJanosch:
work in progress
Camera only has essential methods

### DIFF
--- a/client/src/camera.rs
+++ b/client/src/camera.rs
@@ -46,13 +46,16 @@ impl Camera {
     }
 
     /// Internal function to move the position of the camera
+
     /// Will be called by the other functions (move_forwars etc)
     pub fn move_by(&mut self, pos_diff: Vector3f) {
         self.position += pos_diff;
     }
 
-    /// Method to call when forward movement is needed
+    /// Method to call when **forward movement** is needed
+
     /// `factor` is a factor to scale the movement speed
+
     /// `factor` has to be positive for foward movement
     pub fn move_forward(&mut self, factor: f32) {
         let mut lookatvector = self.get_look_at_vector();
@@ -62,8 +65,10 @@ impl Camera {
         self.move_by(lookatvector);
     }
 
-    /// Method to call when backward movement is needed
+    /// Method to call when **backward movement** is needed
+
     /// `factor` is a factor to scale the movement speed
+
     /// `factor` has to be positive for backward movement
     pub fn move_backward(&mut self, factor: f32) {
         let mut lookatvector = self.get_look_at_vector();
@@ -73,8 +78,10 @@ impl Camera {
         self.move_by(lookatvector);
     }
 
-    /// Method to call when left movement is needed
+    /// Method to call when **left movement** is needed
+
     /// `factor` is a factor to scale the movement speed
+
     /// `factor` has to be positive for left movement
     pub fn move_left(&mut self, factor: f32) {
         let mut lookatvector = self.get_look_at_vector();
@@ -86,8 +93,10 @@ impl Camera {
         self.move_by(move_dir);
     }
 
-    /// Method to call when right movement is needed
+    /// Method to call when **right movement** is needed
+
     /// `factor` is a factor to scale the movement speed
+
     /// `factor` has to be positive for right movement
     pub fn move_right(&mut self, factor: f32) {
         let mut lookatvector = self.get_look_at_vector();
@@ -100,15 +109,19 @@ impl Camera {
     }
 
 
-    /// Method to call when upward movement is needed
+    /// Method to call when **upward movement** is needed
+
     /// `factor` is a factor to scale the movement speed
+
     /// `factor` has to be positive for upward movement
     pub fn move_up(&mut self, factor: f32) {
         self.move_by(Vector3f::new(0.0, 0.0, factor));
     }
 
-    /// Method to call when downward movement is needed
+    /// Method to call when **downward movement** is needed
+
     /// `factor` is a factor to scale the movement speed
+
     /// `factor` has to be positive for downward movement
     pub fn move_down(&mut self, factor: f32) {
         self.move_by(Vector3f::new(0.0, 0.0, -factor));
--- a/client/src/camera.rs
+++ b/client/src/camera.rs
@@ -1,7 +1,12 @@
 use base::math::*;
 
+
 pub struct Camera {
-    // TODO
+    pub position: Point3f, // PRP
+    // look_at_point: Point3<f32>, // VRP
+    pub view_up_vector: Vector3f, // VUV
+    pub theta: f32,
+    pub phi: f32, // theta_phi: (f32, f32),
 }
 
 impl Camera {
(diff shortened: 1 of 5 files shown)


client/src/camera.rs
```diff
@@ -1,7 +1,12 @@
 use base::math::*;
 
+
 pub struct Camera {
-    // TODO
+    pub position: Point3f, // PRP
+    // look_at_point: Point3<f32>, // VRP
+    pub view_up_vector: Vector3f, // VUV
```
LukasKalbertodt: You don't need to save the VUV: it's always just (0,0,1), so it can be created on the fly each time.

client/src/camera.rs
```diff
@@ -1,7 +1,12 @@
...
+    pub theta: f32,
+    pub phi: f32, // theta_phi: (f32, f32),
```
LukasKalbertodt: Please remove the comment

client/src/camera.rs
```diff
@@ -10,10 +15,72 @@ impl Camera {
         perspective(deg(60.0), 16.0 / 9.0, 0.1, 100.0)
     }
 
-    /// Returns view matrix
+    /// Returs view matrix
```
LukasKalbertodt: Why...?

client/src/camera.rs
```diff
@@ -10,10 +15,72 @@ impl Camera {
...
     pub fn view_matrix(&self) -> Matrix4<f32> {
-        Matrix4::look_at(Point3::new(0.0, 0.0, 60.0),
-                         Point3::new(10.0, 10.0, 30.0),
-                         Vector3::unit_z())
+        Matrix4::look_at(self.position, self.get_look_at_point(), self.view_up_vector)
```
LukasKalbertodt: Here you can just keep the `Vector3::unit_z()` as VUV.

client/src/camera.rs
```diff
@@ -10,10 +15,72 @@ impl Camera {
...
+    }
+
+    /// Calculates the look_at_point by using theta and phi on the unit sphere
+    pub fn get_look_at_point(&self) -> Point3f {
+        // let lookatvector = self.get_look_at_vector();
+        self.position + self.get_look_at_vector()
+        // Point3f::new(lookatvector.x + self.position.x,
+        //              lookatvector.y + self.position.y,
+        //              lookatvector.z + self.position.z)
```
LukasKalbertodt: Please remove the inactive code

client/src/camera.rs
```diff
@@ -10,10 +15,72 @@ impl Camera {
...
+    }
+
+    pub fn get_look_at_vector(&self) -> Vector3f {
+        Vector3f::new(self.theta.sin() * self.phi.cos(),
+                      self.theta.sin() * self.phi.sin(),
+                      self.theta.cos())
+    }
+
+    pub fn move_by(&mut self, pos_diff: Vector3f) {
```
LukasKalbertodt: Documentation is missing

client/src/camera.rs
```diff
@@ -10,10 +15,72 @@ impl Camera {
...
+        self.position += pos_diff;
+        // self.position.add_assign(pos_diff);
+    }
```
LukasKalbertodt: commented code again..

LukasKalbertodt:
Documentation should start upper case, so rather write

```rust
/// This is green
fn green() {}
```

than

```rust
/// this is green
fn green() {}
```

client/src/camera.rs
```diff
@@ -10,10 +15,72 @@ impl Camera {
...
+    /// `factor` has to be negative for backward movement
+    pub fn move_forward_backward(&mut self, factor: f32) {
+        // maybe implement the function so the factor is a parameter (to regulate speed
+        // with shift)
+        let mut lookatvector = self.get_look_at_vector();
+        lookatvector.z = 0.0;
+        lookatvector.normalize();
+        lookatvector *= factor;
+        self.move_by(lookatvector); //does the "self" have to be here?
+    }
```
LukasKalbertodt: I wouldn't add this method and choose a different design:

- `move_forward(&mut self, factor: f32)`
- `move_backward(&mut self, factor: f32)`
- `move_left(&mut self, factor: f32)`
- `move_right(&mut self, factor: f32)`

So for now just comment out this whole method.

client/src/camera.rs
```diff
@@ -10,10 +15,72 @@ impl Camera {
...
+    /// `factor` has to be negative for downward movement
+    // pub fn move_up_down(&mut self, factor: f32) {
+    //     self.move_by(Vector3f::new(0.0, 0.0, factor));
+    // }
+
+
+    pub fn change_dir(&mut self, theta_diff: f32, phi_diff: f32) {
+        self.theta += theta_diff;
+        self.phi += phi_diff;
+        // new matrix will be calculated
```
LukasKalbertodt: Documentation is missing. And the last comment is probably wrong: you calculate the matrix on the fly anyway.

client/src/game.rs
```diff
@@ -18,7 +19,13 @@ pub fn run(config: &Config, _: &WorldProvider) -> Result<(), ()> {
...
     let world = World::dummy();
     let world_view = WorldView::from_world(&world, &context);
-    let camera = Camera {};
+    // Initital camera position
+    let camera = Camera {
+        position: Point3::new(0.0, 0.0, 60.0),
+        phi: 0.0,
+        theta: 0.0,
+        view_up_vector: Vector3::unit_z(),
+    };
```
LukasKalbertodt: You can also consider to implement [`Default`](https://doc.rust-lang.org/std/default/trait.Default.html) for this... but not necessary.

client/src/camera.rs
FlorianJanosch: We wanted to keep it, so we might be able to change it in the future (for example for leaning)
(open for discussion, if that feature is needed)

client/src/event_manager.rs
```diff
@@ -23,17 +23,40 @@ impl EventManager {
...
+
         for ev in self.context.poll_events() {
-            match ev {
-                Event::Closed => return EventResponse::Quit,
-                Event::KeyboardInput(_, _, Some(VirtualKeyCode::Escape)) => {
-                    return EventResponse::Quit;
+            for i in 0..handlers.len() {
+                // let x = *handler;
+                // let tmp: &mut _ = &mut **handler;
+                let response = handlers.index_mut(i).handle_event(&ev);
```
jonas-schievink: `handlers[i]` and remove the `ops` import

client/src/ghost.rs
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    pub fn get_camera(&self) -> Camera {
+        self.cam
+    }
+}
+
+impl EventHandler for Ghost {
+    fn handle_event(&mut self, e: &Event) -> EventResponse {
+        match e {
```
jonas-schievink: Prefer `match *e` instead of putting a `&` in front of every pattern

client/src/ghost.rs
```diff
@@ -0,0 +1,59 @@
...
+            &Event::KeyboardInput(ElementState::Pressed, _, Some(VirtualKeyCode::W)) => {
+                self.cam.move_forward(1.0);
+                EventResponse::Continue
+            }
+            &Event::MouseInput(ElementState::Pressed, MouseButton::Left) => {
+                info!("clicked");
+                if let Some(window) = self.context.get_window() {
+                    let res = window.set_cursor_state(CursorState::Grab);
```
jonas-schievink: This is bugged in glium, but I have a workaround ready

client/src/event_manager.rs
```diff
@@ -23,17 +23,40 @@ impl EventManager {
...
+
+pub trait EventHandler {
+    fn handle_event(&mut self, e: &Event) -> EventResponse;
+}
+
+pub struct CloseHandler;
+
+impl EventHandler for CloseHandler {
+    fn handle_event(&mut self, e: &Event) -> EventResponse {
+        match e {
```
jonas-schievink: Prefer `match *e` and remove the `&` in front of the patterns

client/src/game.rs
```diff
@@ -18,12 +20,21 @@ pub fn run(config: &Config, _: &WorldProvider) -> Result<(), ()> {
...
     loop {
-        try!(renderer.render(&world_view, &camera));
+        try!(renderer.render(&world_view, &ghost.get_camera()));
+
+
+        // let mut x: i32 = 3;
+        // let y: &mut i32 = &mut x;
+        // let z: &&mut i32 = &y;
+
+        // let w: &mut i32 = *z;
```
jonas-schievink: Commented code still needs to go

client/src/event_manager.rs
```diff
@@ -23,17 +23,42 @@ impl EventManager {
...
+    pub fn poll_events(&self, mut handlers: Vec<&mut EventHandler>) -> EventResponse {
+
         for ev in self.context.poll_events() {
-            match ev {
-                Event::Closed => return EventResponse::Quit,
-                Event::KeyboardInput(_, _, Some(VirtualKeyCode::Escape)) => {
-                    return EventResponse::Quit;
+            for i in 0..handlers.len() {
+                // let x = *handler;
+                // let tmp: &mut _ = &mut **handler;
```
jonas-schievink: More commented code here

jonas-schievink:
Okay, this is almost ready. The doc comments will look a bit ugly when rendered on the webpage, not sure if you need to fix that. Also needs a rebase.

PR #33, opened 2016-07-21
generate a simple sin-world

adRichter:
additionally creates a few utility methods.

### DIFF
--- a/base/src/gen/world/mod.rs
+++ b/base/src/gen/world/mod.rs
@@ -26,10 +26,11 @@ impl ChunkProvider for WorldGenerator {
         let mut pillars = Vec::new();
         let q = index.0.q * CHUNK_SIZE as i32;
         let r = index.0.r * CHUNK_SIZE as i32;
+        let mut height;
         for i in q..q + CHUNK_SIZE as i32 {
             for j in r..r + CHUNK_SIZE as i32 {
-                pillars.push(HexPillar::from_height(HeightType(((i as f32).sin()
-                    *25.0 + (j as f32).sin()*25.0 + 100.0) as u16)));
+                height = ((i as f32).sin() * 25.0 + (j as f32).sin() * 25.0 + 100.0) as u16;
+                pillars.push(HexPillar::from_height(HeightType(height)));
             }
         }
 
--- a/base/src/gen/world/mod.rs
+++ b/base/src/gen/world/mod.rs
@@ -1,7 +1,7 @@
 //! Procedurally generating the game world.
 //!
 
-use world::{Chunk, ChunkIndex, ChunkProvider};
+use world::{CHUNK_SIZE, Chunk, ChunkIndex, ChunkProvider, HeightType, HexPillar};
 
 /// Main type to generate the game world. Implements the `ChunkProvider` trait
 /// (TODO, see #8).
@@ -22,9 +22,17 @@ impl WorldGenerator {
 }
 
 impl ChunkProvider for WorldGenerator {
-    fn load_chunk(&self, _: ChunkIndex) -> Option<Chunk> {
-        // for the moment returns dummy chunk
-        Some(Chunk::dummy())
+    fn load_chunk(&self, index: ChunkIndex) -> Option<Chunk> {
+        let mut pillars = Vec::<HexPillar>::new();
+        let q = index.0.q * CHUNK_SIZE as i32;
+        let r = index.0.r * CHUNK_SIZE as i32;
+        for i in q..q + CHUNK_SIZE as i32 {
+            for j in r..r + CHUNK_SIZE as i32 {
+                pillars.push(HexPillar::from_height(HeightType(((i as f32).sin()*25.0 + (j as f32).sin()*25.0 + 100.0) as u16)));
+            }
+        }
+
+        Some(Chunk::from_pillars(pillars))
     }
 
     fn is_chunk_loadable(&self, _: ChunkIndex) -> bool {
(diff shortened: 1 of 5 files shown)


base/src/gen/world/mod.rs
```diff
@@ -22,9 +22,17 @@ impl WorldGenerator {
...
-    fn load_chunk(&self, _: ChunkIndex) -> Option<Chunk> {
-        // for the moment returns dummy chunk
-        Some(Chunk::dummy())
+    fn load_chunk(&self, index: ChunkIndex) -> Option<Chunk> {
+        let mut pillars = Vec::<HexPillar>::new();
+        let q = index.0.q * CHUNK_SIZE as i32;
+        let r = index.0.r * CHUNK_SIZE as i32;
+        for i in q..q + CHUNK_SIZE as i32 {
+            for j in r..r + CHUNK_SIZE as i32 {
+                pillars.push(HexPillar::from_height(HeightType(((i as f32).sin()*25.0 + (j as f32).sin()*25.0 + 100.0) as u16)));
```
jonas-schievink: This isn't formatted right, `rustfmt` all the things :D

base/src/gen/world/mod.rs
```diff
@@ -22,9 +22,17 @@ impl WorldGenerator {
 }
 
 impl ChunkProvider for WorldGenerator {
-    fn load_chunk(&self, _: ChunkIndex) -> Option<Chunk> {
-        // for the moment returns dummy chunk
-        Some(Chunk::dummy())
+    fn load_chunk(&self, index: ChunkIndex) -> Option<Chunk> {
+        let mut pillars = Vec::<HexPillar>::new();
```
jonas-schievink: The `::<HexPillar>` shouln't be necessary

base/src/world/chunk.rs
```diff
@@ -28,6 +28,16 @@ impl Chunk {
         Chunk { pillars: pillars }
     }
 
+    pub fn from_pillars(pillars: Vec<HexPillar>) -> Self {
+        // let c: Chunk;
+        if pillars.len() != CHUNK_SIZE.pow(2) as usize {
+            error!("Wrong amount of pillars for chunk!");
+            Chunk::dummy()
```
jonas-schievink: This can probably panic if this invariant is broken. No need to continue when worldgen doesn't work right.

Change it to: `assert_eq!(pillars.len(), CHUNK_SIZE.pow(2));`

base/src/gen/world/mod.rs
jonas-schievink: I suggest you put the `from_height` method into `WorldGenerator` since it's tied to world gen. For example, we'll want to generate props in here, too.

base/src/world/chunk.rs
```diff
@@ -28,14 +28,11 @@ impl Chunk {
         Chunk { pillars: pillars }
     }
 
+    /// Creates a chunk from a Vector<HexPillar>
```
jonas-schievink: `Vec`, not `Vector`. Also, wrap types in backticks (`` ` ``) so they'll be displayed nicer in the docs.

base/src/gen/world/mod.rs
```diff
@@ -22,9 +22,18 @@ impl WorldGenerator {
...
-        // for the moment returns dummy chunk
-        Some(Chunk::dummy())
+    fn load_chunk(&self, index: ChunkIndex) -> Option<Chunk> {
+        let mut pillars = Vec::new();
+        let q = index.0.q * CHUNK_SIZE as i32;
+        let r = index.0.r * CHUNK_SIZE as i32;
+        for i in q..q + CHUNK_SIZE as i32 {
+            for j in r..r + CHUNK_SIZE as i32 {
+                pillars.push(HexPillar::from_height(HeightType(((i as f32).sin()
+                    *25.0 + (j as f32).sin()*25.0 + 100.0) as u16)));
```
LukasKalbertodt: Is this formatted with rustfmt? Looks strange to me...

base/src/world/world.rs
```diff
@@ -36,6 +36,15 @@ impl World {
         }
     }
 
+    pub fn add_chunk(&mut self, index: ChunkIndex, provider: &ChunkProvider) -> Result<(), ()> {
+        if self.chunks.get(&index).is_some() {
```
LukasKalbertodt: Rather use `contains_key()`.

And you can remove the `return` by adding an else branch

client/src/game.rs
```diff
@@ -19,7 +19,7 @@ pub fn run(config: &Config, provider: &ChunkProvider) -> Result<(), ()> {
     let mut world = World::empty();
     for i in 0..3 {
         for j in 0..3 {
-            world.add_chunk(ChunkIndex(AxialPoint::new(i, j)), provider);
+            let _ = world.add_chunk(ChunkIndex(AxialPoint::new(i, j)), provider);
```
LukasKalbertodt: How about at least `warn!()`ing about an `Err` value?

base/src/gen/world/mod.rs
jonas-schievink: It passes Travis... Maybe assign the value inside the `HeightType` to a `let` binding a line before to shorten the line?

client/src/game.rs
jonas-schievink: I'd `unwrap` it. If this fails, something has gone *very* wrong.

LukasKalbertodt: OK, I agree. It's probably not that important anyway, since the whole code is just temporary.

PR #34, opened 2016-07-21
 Rendering of the dummy plant on every hex-pillar

### LINKED ISSUE #12: Basic plant rendering
The dummy plant representation should be rendered to test a few things. Whoever implements this will probably work together with whoever implements #11 (for a few days).

92Andy:
Closes #12

### DIFF
--- a/client/src/world/mod.rs
+++ b/client/src/world/mod.rs
@@ -20,9 +20,9 @@ impl WorldView {
         where F: glium::backend::Facade
     {
         let mut pillars_pos = Vec::new();
-        for q in 0..world::CHUNK_SIZE{
+        for q in 0..world::CHUNK_SIZE {
             for r in 0..world::CHUNK_SIZE {
-                let pos =AxialPoint::new(q.into(), r.into()).to_real();
+                let pos = AxialPoint::new(q.into(), r.into()).to_real();
                 pillars_pos.push(pos);
             }
         }
@@ -31,9 +31,11 @@ impl WorldView {
             chunk: ChunkView::from_chunk(&world.chunks[&ChunkIndex(AxialPoint::new(0, 0))],
                                          AxialPoint::new(0, 0),
                                          facade),
-            plant_view: PlantView::from_dummy_plant(&world.chunks[&ChunkIndex(AxialPoint::new(0, 0))].pillars(),
-                                         pillars_pos,
-                                         facade),
+            plant_view: PlantView::from_dummy_plant(&world.chunks[&ChunkIndex(AxialPoint::new(0,
+                                                                                              0))]
+                                                        .pillars(),
+                                                    pillars_pos,
+                                                    facade),
         }
 
     }
--- a/client/src/world/mod.rs
+++ b/client/src/world/mod.rs
@@ -28,15 +28,14 @@ impl WorldView {
         }
 
         WorldView {
-            chunk: ChunkView::from_chunk(
-                &world.chunks[&ChunkIndex(AxialPoint::new(0, 0))],
-                                        AxialPoint::new(0, 0),
-                                        facade),
-            plant_view: PlantView::from_dummy_plant(
-                &world.chunks[&ChunkIndex(AxialPoint::new(0, 0))]
-                                        .pillars(),
-                                        pillars_pos,
-                                        facade),
+            chunk: ChunkView::from_chunk(&world.chunks[&ChunkIndex(AxialPoint::new(0, 0))],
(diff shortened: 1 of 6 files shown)


client/src/world/mod.rs
```diff
@@ -1,32 +1,48 @@
...
 }
 
 impl WorldView {
     pub fn from_world<F>(world: &World, facade: &F) -> Self
         where F: glium::backend::Facade
     {
+        let mut pillars_pos = Vec::new();
+        for q in 0..world::CHUNK_SIZE{
+            for r in 0..world::CHUNK_SIZE {
+                let pos =AxialPoint::new(q.into(), r.into()).to_real();
```
jonas-schievink: This is badly formatted

client/src/world/mod.rs
```diff
@@ -1,32 +1,48 @@
...
+                pillars_pos.push(pos);
+            }
+        }
+
         WorldView {
             chunk: ChunkView::from_chunk(&world.chunks[&ChunkIndex(AxialPoint::new(0, 0))],
                                          AxialPoint::new(0, 0),
                                          facade),
+            plant_view: PlantView::from_dummy_plant(&world.chunks[&ChunkIndex(AxialPoint::new(0, 0))].pillars(),
```
jonas-schievink: We'll need a vector of `PlantView`s instead of a single one

client/src/world/mod.rs
```diff
@@ -1,32 +1,48 @@
...
+                                         pillars_pos,
+                                         facade),
         }
+
```
jonas-schievink: Unnecessary blank line

client/src/world/plant_view.rs
```diff
@@ -0,0 +1,112 @@
+use glium::backend::Context;
+use Camera;
+use render::ToArr;
+use base::math::*;
+use base::world::{HexPillar, PropType};
+use world::chunk::Vertex;
+use std::rc::Rc;
+use glium;
+
+///Graphical representation of a "base::Plant"
```
jonas-schievink: You should refer to other types using backticks (`` ` ``) instead of `"`

jonas-schievink:
When commenting, always put a space after the `//` (in doc comments, too)

client/src/world/mod.rs
jonas-schievink: (let's do it in a later PR)

client/src/world/plant_view.rs
```diff
@@ -0,0 +1,120 @@
...
+impl PlantView {
+    /// Creates the graphical representation of given plant
+    /// at the given plant offset
+    pub fn from_dummy_plant<F>(hex_vec: &[HexPillar],
+      pillars_pos: Vec<Point2f>, facade: &F) -> Self
+        where F: glium::backend::Facade
+    {
+        /// get height and width for the plant from given Prop
+        let props = hex_vec[0].props();
+        let _prop_ = &props[0];
```
jonas-schievink: Don't use a leading `_` in normal variable names, they'll silence a few compiler warnings (which isn't necessary here)

client/src/world/plant_view.rs
```diff
@@ -0,0 +1,120 @@
...
+    program: glium::Program,
+    plants_pos: Vec<Point2f>,
+    /// Context for IndexBuffer
+    con: Rc<Context>,
+}
+
+impl PlantView {
+    /// Creates the graphical representation of given plant
+    /// at the given plant offset
+    pub fn from_dummy_plant<F>(hex_vec: &[HexPillar],
```
jonas-schievink: This function should really get the `Plant` as an argument instead of matching the prop itself (that's something a `PropView` should do, delegating the plant case to the `PlantView` - but we'll do that later). Probably can land this without this change though.

client/src/world/plant_view.rs
```diff
@@ -0,0 +1,120 @@
...
+      pillars_pos: Vec<Point2f>, facade: &F) -> Self
+        where F: glium::backend::Facade
+    {
+        /// get height and width for the plant from given Prop
+        let props = hex_vec[0].props();
+        let _prop_ = &props[0];
+        let p_type = &_prop_.prop;
+        let width: f32;
+        let height: f32;
+        match p_type {
```
jonas-schievink: `let (width, height) = match hex_vec[0].props()[0].prop { ... };`

jonas-schievink: `hex_vec[0]` is wrong since it'll always create a view for the plant in the first hex pillar

PR #39, opened 2016-07-22
Add get method to chunk.rs

### LINKED ISSUE #10: Non-panicking index method for `Chunk`
Chunk is indexable with `ops::Index`, but this panics if the position is out of bounds. There should be a safer method which returns an `Option<_>`.

JonasKnerr:
closes #10

### DIFF
--- a/base/src/world/chunk.rs
+++ b/base/src/world/chunk.rs
@@ -47,9 +47,10 @@ impl ops::Index<AxialPoint> for Chunk {
     type Output = HexPillar;
 
     fn index(&self, pos: AxialPoint) -> &Self::Output {
-        let chunk_size: PillarIndexComponent = CHUNK_SIZE.into();
-
-        assert!(pos.q >= 0 && pos.q < chunk_size && pos.r >= 0 && pos.r < chunk_size);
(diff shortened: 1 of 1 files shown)


base/src/world/chunk.rs
```diff
@@ -31,17 +31,25 @@ impl Chunk {
...
 
     fn index(&self, pos: AxialPoint) -> &Self::Output {
         let chunk_size: PillarIndexComponent = CHUNK_SIZE.into();
-        assert!(pos.q >= 0 && pos.q < chunk_size && pos.r >= 0 && pos.r < chunk_size,
-                "axial position to index `Chunk` are out of bounds: {:?}",
-                pos);
 
-        &self.pillars[(pos.r as usize) * (CHUNK_SIZE as usize) + (pos.q as usize)]
+        assert!(pos.q >= 0 && pos.q < chunk_size && pos.r >= 0 && pos.r < chunk_size);
+        (&self.pillars[(pos.r as usize) * (CHUNK_SIZE as usize) + (pos.q as usize)])
```
jonas-schievink: Why change this? Wasn't this good before? I suppose you could call the new `get` method and unwrap it's result with `unwrap_or_else` (consult the std docs for how to do that)

PR #41, opened 2016-07-22
Finalized Chunk Rendering

NilsAffing:


client/src/world/chunk.rs
```diff
@@ -56,12 +56,17 @@ impl ChunkView {
...
-                positions.push(pos);
-            }
+        let mut pillars = Vec::new();
+        for q in 0..world::CHUNK_SIZE * world::CHUNK_SIZE {
+            let pos = offset.to_real() +
+                      AxialVector::new((q / world::CHUNK_SIZE).into(),
+                                       (q % world::CHUNK_SIZE).into())
+                .to_real();
+            pillars.push(PillarView::from_pillar_section(pos,
+                                                         _chunk.pillars()
```
LukasKalbertodt: Please rename this variable... variable names should never begin with underscore.

client/src/world/chunk.rs
```diff
@@ -79,42 +84,44 @@ impl ChunkView {
...
-            let params = glium::DrawParameters {
-                depth: glium::Depth {
-                    write: true,
-                    test: glium::draw_parameters::DepthTest::IfLess,
+        for pillar in &self.pillars {
+            for section in &pillar.sections {
+                let uniforms = uniform!{
+                  scale_matrix:[[1.0,0.0,0.0,0.0]
+                               ,[0.0,1.0,0.0,0.0],
+                                 //FIXME 1.0 heightvalue
```
LukasKalbertodt: This FIXME can be removed, right?

client/src/world/chunk.rs
```diff
@@ -127,3 +134,29 @@ struct Vertex {
...
+    sections: Vec<[f32; 2]>,
+}
+
+impl PillarView {
+    // fn new(pos: Point2f, bottom: HeightType, height: HeightType) -> PillarView {
+    // PillarView {
+    // pos: pos,
+    // sections: vec![bottom.0 as f32,height.0 as f32],
+    // }
+    // }
```
LukasKalbertodt: If you don't need that, remove the code completely

client/src/world/chunk.rs
```diff
@@ -127,3 +134,29 @@ struct Vertex {
 }
 
 implement_vertex!(Vertex, position, color);
+
+pub struct PillarView {
+    pos: Point2f,
+    sections: Vec<[f32; 2]>,
```
LukasKalbertodt: Why don't you just save a `Vec<PillarSection>`? Then you can save some code to avoid converting and you need the material from the `PillarSection` later on anyway.

NilsAffing: This will be changed soon, because we are working on this data structure.

PR #42, opened 2016-07-22
Camera and EventHandler Fix for MouseMovement

FlorianJanosch:
fixed MouseMovement for left and right (it was the other way around)


jonas-schievink:
A few warnings need to be fixed

PR #46, opened 2016-07-22
Server preparations

jonas-schievink:
This adds a dummy server, starts it with the client, and connects to it.

### DIFF
--- a/base/src/msg.rs
+++ b/base/src/msg.rs
@@ -4,7 +4,7 @@ use math::{Point3f, Vector3f};
 
 /// A message from the server to a client.
 #[derive(RustcEncodable, RustcDecodable)]
-pub enum ClientCommand {
+pub enum ServerMessage {
     /// Register a player currently playing on the same server.
     ///
     /// Sent when a player joins the server or when this client just joined the
@@ -16,7 +16,7 @@ pub enum ClientCommand {
 
 /// A message from a client to the server.
 #[derive(RustcEncodable, RustcDecodable)]
-pub enum ServerCommand {
+pub enum ClientMessage {
     UpdatePose {
         position: Point3f,
         orientation: Vector3f,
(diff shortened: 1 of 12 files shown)


server/src/server.rs
```diff
@@ -0,0 +1,94 @@
...
+                    Err(TryRecvError::Empty) => {}
+                    Err(TryRecvError::Disconnected) => {
+                        info!("tcp listener thread exited, killing server");
+                        return Ok(());
+                    }
+                }
+            }
+
+            // self.world_manager.update_world(self.player.get_camera().position);
+        }
```
LukasKalbertodt: Two `loop`s? For extra infinity :wink:  And shouldn't we at least sleep for 50ms instead of doing super busy waiting?

base/src/msg.rs
```diff
@@ -0,0 +1,24 @@
...
+}
+
+/// A message from a client to the server.
+#[derive(RustcEncodable, RustcDecodable)]
+pub enum ServerCommand {
+    UpdatePose {
+        position: Point3f,
+        orientation: Vector3f,
+    },
+}
```
LukasKalbertodt: As discussed: rename to `ServerMessage` and `ClientMessage`

server/src/server.rs
jonas-schievink: The outer loop is supposed to do world updates. I'll put in a rate limit. I want the server to process all messages once per loop, that's what the inner loop is for.

base/src/msg.rs
```diff
@@ -16,7 +16,7 @@ pub enum ClientCommand {
 
 /// A message from a client to the server.
 #[derive(RustcEncodable, RustcDecodable)]
-pub enum ServerCommand {
+pub enum ClientMessage {
     UpdatePose {
```
LukasKalbertodt: Oh... is `Pose` intentionally? Shouldn't it be `Pos`?

jonas-schievink: "Pose" as in "posture", meaning position and orientation in 3D space.

PR #35, opened 2016-07-21
Add to_arr

JonasKnerr:
closes #17

PR #52, opened 2016-07-22
Temp. plant renderer and treegen fixes/improvements

jonas-schievink:
This is ready to go now!


base/src/gen/mod.rs
```diff
@@ -1,8 +1,44 @@
...
+fn seeded_rng<T: Hash, U: Hash>(world_seed: u64, feat_seed: T, loc_seed: U) -> Random {
+    // Hash everything, even `world_seed`, since XorShift really doesn't like seeds
+    // with many 0s in it
+    let mut fnv = FnvHasher::default();
+    world_seed.hash(&mut fnv);
+    feat_seed.hash(&mut fnv);
+    loc_seed.hash(&mut fnv);
+    let rng_seed = fnv.finish();
+    let seed0 = (rng_seed >> 32) as u32;
+    let seed1 = rng_seed as u32;
```
LukasKalbertodt: I hate this `as` cast more every day. It would be way nicer to express your intend clearly in the code, like `let seed1: u32 = rng_seed.truncate();`. Ok but I guess this can't be improved right now :disappointed:

base/src/gen/mod.rs
```diff
@@ -1,8 +1,44 @@
...
+    // Apply our patented cryptoperturbation mask to magically extend the 64-bit
+    // hash to 128 bits used by xorshift:
+    // (To be honest, I just didn't want to pass the same 2 word twice)
+    let seed2 = seed0 ^ 0xdeadbeef;
+    let seed3 = seed1 ^ 0xcafebabe;
```
LukasKalbertodt: I'm not sure about this. This obviously effects the world we are generating for a given seed, so changing this hacky solution is not backwards compatible. Furthermore we might want to use another RNG in the future, which is seeded by a different number of bytes.

Masking our hashed seed like that probably doesn't increase the quality of our random numbers at all :stuck_out_tongue: So I guess we should just repeat our hashed seed as many times as necessary to completely seed the RNG.

base/src/gen/plant/tree.rs
```diff
@@ -169,17 +177,23 @@ impl TreeGen {
...
         let trunk_height = range_sample(&self.preset.trunk_height, rng);
         let trunk_diameter_top = range_sample(&self.preset.trunk_diameter_top, rng);
-        let min_branch_height = range_sample(&self.preset.min_branch_height, rng);
+        let min_branch_height = range_sample(&self.preset.min_branch_height, rng) * trunk_height;
+
+        info!("trunk diam {} to {}, height {}, branch start at {}",
+              trunk_diameter,
+              trunk_diameter_top,
+              trunk_height,
+              min_branch_height);
```
LukasKalbertodt: Shouldn't this be `debug!()`?

base/src/gen/plant/tree.rs
```diff
@@ -194,19 +208,18 @@ impl TreeGen {
...
-            let mut height = 0.0;
-            while height < trunk_height {
-                // Current height as a fraction of the total height
-                let height_frac = if height == 0.0 { 0.0 } else { trunk_height / height };
+
+            // Split trunk in segments
+            // FIXME Vary the segment direction like we do for normal branches
+            // FIXME Make segment count depend on the trunk height
+            const SEGMENT_COUNT: u32 = 10;
+            for i in 0..SEGMENT_COUNT + 1 {
```
LukasKalbertodt: C'mon https://github.com/rust-lang/rust/issues/28237 *____*

base/src/gen/world/mod.rs
```diff
@@ -26,12 +29,29 @@ impl ChunkProvider for WorldGenerator {
...
+                const TREE_SPACING: i32 = 8;
+                if i % TREE_SPACING == 0 && j % TREE_SPACING == 0 {
+                    let mut rng = super::seeded_rng(self.seed, "TREE", (i, j));
+                    let gen = PlantGenerator::rand(&mut rng);
+
+                    props.push(Prop {
+                        baseline: HeightType(height),
+                        prop: PropType::Plant(gen.generate(&mut rng)),
+                    });
+                }
```
LukasKalbertodt: Maybe if-else to avoid mutable vector?

client/src/world/chunk_view.rs
```diff
@@ -1,18 +1,23 @@
-use base::world::{self, Chunk, PillarSection};
+use base::world::{self, Chunk, HexPillar, PillarSection, PropType};
 use base::math::*;
-use glium;
+use glium::{Depth, DrawParameters, IndexBuffer, Program, Surface, VertexBuffer};
```
LukasKalbertodt: Mh... `Surface`, `Program` and `Depth` are rather ambiguous without `glium`, right? I'd prefer writing `glium::Program` in the code... but I guess we can keep it that way.

client/src/world/chunk_view.rs
```diff
@@ -139,17 +137,25 @@ implement_vertex!(Vertex, position, color);
 pub struct PillarView {
     pos: Point2f,
     pillars: Vec<PillarSection>,
```
LukasKalbertodt: I know, you did not invent this name, but this should really be called `pillar_sections`, otherwise it's super confusing :astonished:

client/src/world/plant_view.rs
```diff
@@ -1,111 +1,83 @@
-use glium::backend::Context;
+use glium::backend::Facade;
+use glium::{Depth, DepthTest, DrawParameters, Program, Surface, VertexBuffer};
```
LukasKalbertodt: Same as above.

client/src/world/plant_view.rs
```diff
@@ -1,111 +1,83 @@
...
-    /// at the given plant offset
-    pub fn from_dummy_plant<F>(hex_vec: &[HexPillar], pillars_pos: Vec<Point2f>, facade: &F) -> Self
-        where F: glium::backend::Facade
-    {
-        // get height and width for the plant from given Prop
-        let (width, height) = match hex_vec[0].props()[0].prop {
-            PropType::Plant(p) => (p.stem_width, p.height),
-        };
+    pub fn from_plant<F: Facade>(pos: Point3f, plant: &Plant, facade: &F) -> Self {
+        // FIXME this is just stupid, don't recompile the shader for every plant
```
LukasKalbertodt: Maybe the helper from #65 should cache shader and programs?

client/src/world/plant_view.rs
```diff
@@ -1,111 +1,83 @@
...
+        let mut verts = 0;
+        let branches = match *plant {
+            Plant::Tree { ref branches } => {
+                branches.iter()
+                    .map(|branch| {
+                        let mut vertices = Vec::new();
+                        for cp in branch.points.iter() {
+                            verts += 1;
+                            vertices.push(Vertex {
+                                position: [cp.point.x, cp.point.y, cp.point.z],
```
LukasKalbertodt: With #35 you should be able to say `position: cp.point.to_arr()`

client/src/world/plant_view.rs
```diff
@@ -1,111 +1,83 @@
...
+                                color: [branch.color.x, branch.color.y, branch.color.z],
```
LukasKalbertodt: I missed that in your last PR, but: we should maybe use a proper type for storing colors?

LukasKalbertodt:
I made a few line comments. Otherwise this PR looks nice. I just checked it out to see it in action ... neat :open_mouth:

client/src/world/plant_view.rs
jonas-schievink: We should really pass a `Renderer` around that does this for us (or something similar)

base/src/gen/world/mod.rs
jonas-schievink: My intention was that other world features could be added after trees. But we'll probably use a completely different and more flexible system then.

client/src/world/chunk_view.rs
jonas-schievink: I don't find `Surface` and `Program` ambiguous, what other use of these names is there? `Depth` seems kinda ambiguous, true.

client/src/world/plant_view.rs
jonas-schievink: Seems like that was only implemented for `Vector3`, not for `Point3`. I'll impl it myself.

client/src/world/chunk_view.rs
LukasKalbertodt: Surface of the stem? Surface of a leaf? Ok, `Program` will probably not occur in any other form in our game...

base/src/gen/mod.rs
jonas-schievink: > Furthermore we might want to use another RNG in the future, which is seeded by a different number of bytes.

...which would be incompatible, too :P

LukasKalbertodt: Oh... right ^_^ 

I created #66 to discuss that someday in the future.

PR #67, opened 2016-07-23
Treegen and rendering tweaks/fixes and cleanup

jonas-schievink:
* `debug!` the config we use
* Fix growing of new tree branches
 * Their growing direction wasn't randomized properly
* Higher chance of growing branches out of the trunk
* Stop extending branches when they have a diameter of <1mm
* Add an FPS counter which prints the FPS to the console every second
* Make `HeightType` a tiny bit less confusing
* Chunk rendering: Drop the `scale_matrix` and pass a `height` instead
* Make material colors RGB instead of RGBA
 * The shader ignored the Alpha channel anyways
 * I'm not convinced that we need or want transparent ground

### DIFF
--- a/base/src/gen/plant/tree.rs
+++ b/base/src/gen/plant/tree.rs
@@ -173,7 +173,7 @@ impl TreeGen {
         // calculate the cross product between it and `parent_dir`.
         // This gets us a vector perpendicular to `parent_dir`, which we can rotate
         // `parent_dir` around to tilt it by `angle` degrees.
-        let mut rand_vec = Vector3f::rand(rng).normalize();
+        let mut rand_vec = Vector3f::rand(rng);
         while rand_vec.angle(parent_dir).approx_eq(&Rad::new(0.0)) {
             // They're parallel, so generate a new vector. This is probably unnecessary,
             // but someone
@@ -185,7 +185,7 @@ impl TreeGen {
         let rot_vec = parent_dir.cross(rand_vec);
 
         // Tilt the growing direction of the parent branch by about 90° to get the
-        // direction of the
+        // direction of the new branch
         let tilt_rotation = Basis3::from_axis_angle(rot_vec, Deg::new(angle).into());
 
         let spin_angle = range_sample(&(0.0..360.0), rng);
--- a/base/src/gen/plant/tree.rs
+++ b/base/src/gen/plant/tree.rs
@@ -1,7 +1,8 @@
 //! Generates random trees and tree-like plants.
 
 use prop::plant::{Branch, ControlPoint};
-use math::{Basis3, Deg, Euler, InnerSpace, Point3f, Rotation, Vector1, Vector3f};
+use math::{ApproxEq, Basis3, Deg, Euler, InnerSpace, Point3f, Rad, Rotation, Rotation3, Vector1,
+           Vector3f};
 use rand::{Rand, Rng};
 use rand::distributions::range::SampleRange;
 use rand::distributions::{self, IndependentSample};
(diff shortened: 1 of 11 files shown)


base/src/gen/plant/tree.rs
```diff
@@ -1,7 +1,8 @@
 //! Generates random trees and tree-like plants.
 
 use prop::plant::{Branch, ControlPoint};
-use math::{Basis3, Deg, Euler, InnerSpace, Point3f, Rotation, Vector1, Vector3f};
+use math::{ApproxEq, Basis3, Deg, Euler, InnerSpace, Point3f, Rad, Rotation, Rotation3, Vector1,
+           Vector3f};
```
LukasKalbertodt: Nearly everywhere else we `use math::*;`. I think it's OK to use wildcard imports for `math` and `math` alone.

jonas-schievink: https://osnacs.github.io/plantex/cgmath/trait.PartialOrd.html conflicts with the `std` trait of the same name /:

I explicitly named it via `cmp::PartialOrd` to work around this

base/src/gen/plant/tree.rs
```diff
@@ -157,20 +167,33 @@ impl TreeGen {
...
+        // AWESOME hack: Generate *any* vector that is not parallel to `parent_dir`,
+        // calculate the cross product between it and `parent_dir`.
+        // This gets us a vector perpendicular to `parent_dir`, which we can rotate
+        // `parent_dir` around to tilt it by `angle` degrees.
+        let mut rand_vec = Vector3f::rand(rng).normalize();
+        while rand_vec.angle(parent_dir).approx_eq(&Rad::new(0.0)) {
+            // They're parallel, so generate a new vector. This is probably unnecessary,
+            // but someone
+            // who *actually* knows their math instead of faking it should check.
+            rand_vec = Vector3f::rand(rng);
```
LukasKalbertodt: This vector should be normalized as well, if you already normalize it 5 lines above. It doesn't need to be normalized, but normalizing it will increase floating point accuracy a bit, I think.

I would also suggest not passing `Rad::new(0.0)`, but a very small non-zero number. If the angle between the two vectors is very close to zero, the floating point math for calculating the cross product and using it will be rather inaccurate.

base/src/gen/plant/tree.rs
```diff
@@ -157,20 +167,33 @@ impl TreeGen {
...
+        }
+
+        // Create vector around we'll tilt `parent_dir`
+        let rot_vec = parent_dir.cross(rand_vec);
+
+        // Tilt the growing direction of the parent branch by about 90° to get the
+        // direction of the
```
LukasKalbertodt: Unfinished comment?

base/src/gen/plant/tree.rs
jonas-schievink: I only normalized it above because I wasn't sure what would happen if the random vector ends up extremely large.

The `approx_eq` should in theory already use such a threshold.

base/src/gen/plant/tree.rs
```diff
@@ -157,20 +167,33 @@ impl TreeGen {
...
+        let tilt_rotation = Basis3::from_axis_angle(rot_vec, Deg::new(angle).into());
+
+        let spin_angle = range_sample(&(0.0..360.0), rng);
+
+        // Then rotate this vector randomly (0-360°) around the parent branch
+        let around_we_go = Basis3::from_axis_angle(parent_dir, Deg::new(spin_angle).into());
+
+        around_we_go.rotate_vector(tilt_rotation.rotate_vector(parent_dir))
```
LukasKalbertodt: To get `tilt_rotation` we already generate a random vector that is not parallel to `parent_dir`. So it shouldn't be necessary to rotate the tilted vector by a random amount again, right?

jonas-schievink: `Vector3f::rand` will return a vector with all elements in the range `[0..1)`... apparently. So when I don't do this the trees look worse since all branches point in roughly the same direction.

base/src/gen/plant/tree.rs
jonas-schievink: Okay since `rand_vec` will only have elements in range `[0..1)` (see comment below), the `.normalize()` is unnecessary.

PR #70, opened 2016-07-24
Remove `config.toml`: it's user specfic and shouldn't be in git

LukasKalbertodt:


jonas-schievink:
Maybe also put it in the `.gitignore` and ship a `config.toml.example` instead?

PR #73, opened 2016-07-25
Implement Delta

FlorianJanosch:
delta now is paramter for update function of ghost
delta will now influence the speed of the player, so the player still moves with sufficient speed, even though the game might have low FPS

### DIFF
--- a/client/src/game.rs
+++ b/client/src/game.rs
@@ -39,8 +39,17 @@ impl Game {
     pub fn run(mut self) -> Result<(), ()> {
         let mut frames = 0;
         let mut next_fps_measure = Instant::now() + Duration::from_secs(1);
+        let mut time_prev = Instant::now();
         loop {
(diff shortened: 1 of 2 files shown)


client/src/game.rs
```diff
@@ -39,16 +39,25 @@ impl Game {
...
         let mut next_fps_measure = Instant::now() + Duration::from_secs(1);
+        let mut time_prev = Instant::now();
         loop {
             self.world_manager.update_world();
 
+
+            let mut time_now = Instant::now();
+            let mut duration_delta = time_now.duration_since(time_prev);
+            let mut delta = ((duration_delta.subsec_nanos() / 1000) as f32) / 1000000.0 +
+                            duration_delta.as_secs() as f32;
```
LukasKalbertodt: To increase readability, number literals can use underscores. So please write `1_000_000.0` instead :)

PR #74, opened 2016-07-25
Close #68

### LINKED ISSUE #68: Implement AxialPoint::from_real
We need to convert world coordinates to axial coordinates to be able to figure out the player position (or any object's position) in pillar/chunk coordinates to implement things like #55, so this has very high priority at least for world gen.

http://www.redblobgames.com/grids/hexagons/#pixel-to-hex

`pub fn from_real(real: Point2f) -> Self`

(this doesn't have much to do with world gen, but I tagged it `S-gen-world` anyways because it's basically blocked on this)

92Andy:


base/src/math/axial_point.rs
```diff
@@ -34,6 +34,30 @@ impl AxialPoint {
             y: (self.r as DefaultFloat) * (3.0 / 2.0) * HEX_OUTER_RADIUS,
         }
     }
+    /// Return the AxailPoint from a Point2f
```
jonas-schievink: `AxialPoint`

base/src/math/axial_point.rs
```diff
@@ -34,6 +35,32 @@ impl AxialPoint {
             y: (self.r as DefaultFloat) * (3.0 / 2.0) * HEX_OUTER_RADIUS,
         }
     }
+
+    /// Return the `AxailPoint´ from a Point2f
```
jonas-schievink: Typo: AxailPoint should be AxialPoint´

jonas-schievink: And `Point2f` in backticks `` ` ``

PR #75, opened 2016-07-25
Closure for iterating through a Chunk. Close #63

### LINKED ISSUE #63: Fancy chunk initialization method
Currently at many places in our code we have two nested for looks like this:

```rust
for q in 0..CHUNK_SIZE {
    for r in 0..CHUNK_SIZE { ... }
}
```

I want to replace this by

```rust
Chunk::with_pillars(|q, r| ... )
```

EDIT: I guess the closure should get one `AxialPoint` as parameter instead of q and r...

92Andy:


base/src/world/chunk.rs
```diff
@@ -39,6 +39,18 @@ impl Chunk {
             Some(&self.pillars[(pos.r as usize) * (CHUNK_SIZE as usize) + (pos.q as usize)])
         }
     }
+
+    /// Closure function which interates through a `Chunk`
```
jonas-schievink: Maybe something like this would be better (since there's no such thing as "closure functions"):

"Calls the given closure with all pillar positions that are contained in every chunk"

jonas-schievink: or "in a chunk"

base/src/world/chunk.rs
```diff
@@ -51,6 +51,19 @@ impl Chunk {
             }
         }
     }
+
+    pub fn with_pillars<F>(mut iter: F) -> Chunk
```
jonas-schievink: Documentation missing. Maybe this:

"Creates a chunk using individual pillars returned by a closure"

PR #76, opened 2016-07-25
Dynamic chunk lagging

### LINKED ISSUE #55: Dynamic chunk loading and unloading
Currently, a fixed number of chunks at a fixed position is loaded. We need a system that dynamically loads and unloads chunks depending on the position of the player/ghost. This unloading probably only needs to happen for `ChunkView`s ... the `Chunk`s don't occupy a lot of RAM and aren't drawn without corresponding `ChunkView`, so they can stay. Of course if a chunk is not even generated as a `Chunk`, this needs to happen, too.

Additionally there should be some clever distance based measure to unload/load chunks. While it is easier to always load neighbor chunks, the strange shape of our chunks makes it rather inefficient. Two corners of the rhombus are very far away from the player (we don't really need to load them) while the other two corners are pretty close. So the algorithm should load every chunk which is less than n meters away, measured in world coordinates.

jonas-schievink:
Closes #55 

🎊

### DIFF
--- a/client/src/game.rs
+++ b/client/src/game.rs
@@ -21,14 +21,11 @@ impl Game {
         let facade = try!(create_context(&config));
         let context = Rc::new(GameContext::new(facade, config.clone()));
 
-        let world_manager = WorldManager::new(create_chunk_provider(context.get_config()),
-                                              context.clone());
-        world_manager.pregenerate_world();
-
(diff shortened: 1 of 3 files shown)


jonas-schievink:
(no longer lags that badly)

Issue #50, opened 2016-07-22
Tracking issue: world generation


LukasKalbertodt:
This is the tracking issue for world generation -- it's not expected to be closed anytime soon. We have a few goals we'd like to reach if possible:

- [x] rough-grained noise maps for for various properties of the terrain
  - [x] temperature
  - [x] humidity
  - [x] type of terrain height variations (hills, mountains, valleys, meadows, ...)
- [x] the height of the terrain
- [x] material of the terrain
- [x] generation of plants
- [ ] rivers
- [ ] caves

(list may be expaned)

### Useful links:

- [OpenSimplex as open noise function](https://www.reddit.com/r/gamedev/comments/2gw6wz/like_perlins_simplex_noise_but_dont_like_the/)
- [`worldgen` crate](https://crates.io/crates/worldgen)

jonas-schievink:
There's already an OpenSimplex noise implementation in Rust: https://github.com/bjz/noise-rs (sadly, no non-broken docs are hosted anywhere)

jonas-schievink:
Temporary docs for the `noise` crate: https://jonas-schievink.github.io/noise-rs/noise/index.html

The API isn't as nice as the `worldgen` crate API, but if it works it works.

LukasKalbertodt:
We could implement the [`NoiseProvider` trait](http://www.samuelsleight.co.uk/rust-docs/worldgen/worldgen/noise/trait.NoiseProvider.html) for a wrapper type around `OpenSimplex2`, right? This would allow us to use the `worldgen` API. It's a bit sad that `worldgen` is limited to 2D noise functions. If we want to use `OpenSimplex3` later, we had to use the `noise` API anyway...

jonas-schievink:
So apparently `worldgen` requires nightly Rust because it uses some syntax extension. Would be really nice if that was more obvious to see (other than compilation failing). Gonna use `noise` without it then...

toschneider:
object placement http://gamedev.stackexchange.com/questions/31724/how-can-you-procedurally-place-objects-in-a-non-gridded-game

jonas-schievink:
A nice biome diagram: http://w3.marietta.edu/%7Ebiol/biomes/biome_main.htm

jonas-schievink:
Okay so I'm playing around with using 3D noise for terrain generation instead of a 2D height map, and it's looking extremely interesting (unlike our 2D terrain, which was kinda boring).

LukasKalbertodt:
[...]

PR #77, opened 2016-07-25
add terrain generation with open simplex noise

KarolinePlum:
progress on issue #50

### DIFF
--- a/base/src/gen/world/mod.rs
+++ b/base/src/gen/world/mod.rs
@@ -3,7 +3,7 @@
 
 // extern crate noise;
 
-use world::{CHUNK_SIZE, Chunk, ChunkIndex, ChunkProvider, HeightType, HexPillar};
+use world::{Chunk, ChunkIndex, ChunkProvider, HeightType, HexPillar};
 use world::{GroundMaterial, PillarSection, Prop, PropType};
 use rand::Rand;
 use gen::PlantGenerator;
@@ -25,11 +25,12 @@ pub struct WorldGenerator {
 impl WorldGenerator {
     /// Creates the generator with the given seed.
     pub fn with_seed(seed: u64) -> Self {
-        let mut rng = seeded_rng(seed, 0, ());
+        let mut terrain_rng = seeded_rng(seed, 0, ());
 
         WorldGenerator {
             seed: seed,
-            terrain_table: PermutationTable::rand(&mut rng),
+            terrain_table: PermutationTable::rand(&mut terrain_rng),
+            plant_table: PermutationTable::rand(&mut plant_rng),
         }
     }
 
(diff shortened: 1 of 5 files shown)


base/src/gen/world/mod.rs
```diff
@@ -1,21 +1,36 @@
...
 use world::{CHUNK_SIZE, Chunk, ChunkIndex, ChunkProvider, HeightType, HexPillar};
 use world::{GroundMaterial, PillarSection, Prop, PropType};
 use rand::Rand;
 use gen::PlantGenerator;
+use noise::{PermutationTable, open_simplex2};
+use gen::seeded_rng;
+
+// pub const LANDSCAPE_FLAT: f32 = 0.01;
+pub const LANDSCAPE_MODERATE: f32 = 0.05;
+// pub const LANDSCAPE_HILLY: f32 = 0.1125;
```
LukasKalbertodt: Do you need those constants later? If not, please remove the comments ;)

base/src/gen/world/mod.rs
```diff
@@ -32,12 +47,15 @@ impl ChunkProvider for WorldGenerator {
...
 
+                let noise = open_simplex2::<f32>(&self.terrain_table,
+                                                 &[(i as f32) * LANDSCAPE_MODERATE,
+                                                   (j as f32) * LANDSCAPE_MODERATE]);
+                let height = 50.0 + noise * 50.0;
+                println!("{}", height);
                 let ground_section = PillarSection::new(GroundMaterial::Dirt,
                                                         HeightType::from_units(0),
-                                                        HeightType::from_units(height));
+                                                        HeightType::from_units(height as u16));
```
LukasKalbertodt: I think you can use `height.into()` here, right?

base/src/gen/world/mod.rs
```diff
@@ -47,7 +65,7 @@ impl ChunkProvider for WorldGenerator {
                     let gen = PlantGenerator::rand(&mut rng);
 
                     props.push(Prop {
-                        baseline: HeightType::from_units(height),
+                        baseline: HeightType::from_units(height as u16),
```
LukasKalbertodt: Same as above.

base/src/gen/world/mod.rs
KarolinePlum: Yes we might need them later for generating biomes

base/src/gen/world/mod.rs
```diff
@@ -1,21 +1,36 @@
 //! Procedurally generating the game world.
 //!
 
-use world::{CHUNK_SIZE, Chunk, ChunkIndex, ChunkProvider, HeightType, HexPillar};
+// extern crate noise;
```
jonas-schievink: Unnecessary comment

base/src/world/chunk.rs
```diff
@@ -54,12 +54,15 @@ impl Chunk {
     }
 
     /// Creates a `Chunk` using individual pillars returned by a closure
-    pub fn with_pillars<F>(mut func: F) -> Chunk
+    pub fn with_pillars<F>(chunk_index: ChunkIndex, mut func: F) -> Chunk
```
jonas-schievink: I'm fine with this personally, but you could also do this calculation in the closure you pass to this

PR #80, opened 2016-07-25
Fancy iterator over chunks and pillars. Close #64

### LINKED ISSUE #64: Fancy iterators over chunks and pillars
Currently, we can use `Chunk::pillars()` to get a slice over all pillars, but this is hardly used, because the pillar itself don't know where it's positioned in the world. Therefore we often have two nested for loops in user code.

I want a `pillars()` method that returns an iterator over the item `(AxialVector, &HexPillar)`. A vector instead of a point, because it's not the world position. Or maybe `PillarPosition` ... I lost track of all axial position types we have :stuck_out_tongue:  

The same applied to the `World` type: we maybe want to iterate over all chunks while also knowing the chunk position.

92Andy:


client/src/world/chunk_view.rs
```diff
@@ -138,9 +144,8 @@ impl PillarView {
     fn from_pillar<F: Facade>(pos: Point2f, pillar: &HexPillar, facade: &F) -> PillarView {
         PillarView {
             pos: pos,
-            sections: pillar.sections().to_vec(),
+            sections: pillar.sections().as_slice().to_vec(),
```
jonas-schievink: `.sections().collect()`: https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.collect

client/src/world/chunk_view.rs
```diff
@@ -60,7 +60,13 @@ impl ChunkView {
                       AxialVector::new((q / world::CHUNK_SIZE).into(),
                                        (q % world::CHUNK_SIZE).into())
                 .to_real();
-            let pillar = &chunk.pillars()[q as usize];
+
+            // Every element is reachable because `Chunk` generates
+            // a filled vector of `HexPillar` which changed in an Iterator
+            let pillar = match chunk.pillars().next() {
```
jonas-schievink: This isn't equivalent to the code on the left: The left side takes the element `q`, you return element `0` (the first one).

base/src/world/chunk.rs
```diff
@@ -26,8 +27,8 @@ impl Chunk {
         Chunk { pillars: pillars }
     }
 
-    pub fn pillars(&self) -> &[HexPillar] {
-        &self.pillars
+    pub fn pillars(&self) -> Iter<HexPillar> {
```
jonas-schievink: This is supposed to iterate over tuples of `(AxialVector, &HexPillar)`

PR #84, opened 2016-07-25
Move `PlantView` into `view` module and add `PlantRenderer`

LukasKalbertodt:


jonas-schievink:
LGTM. This blocks some performance optimizations we really need, so I'll merge now.

PR #86, opened 2016-07-25
Introduce a ChunkRenderer and render chunks via Instancing

### LINKED ISSUE #87: Implement instancing for pillar rendering
We postponed #61 and decided to use instancing to improve performance instead.

jonas-schievink:
Speeds up world rendering a bit since we don't use a different program
(and vertex buffer) object for each drawn chunk.

Instancing multiplies my FPS by about ~7

Closes #87


jonas-schievink:
There's also a high chance that this gets rid of the absurd lag spikes that occur when new chunks are loaded (I'll check it, gimme a sec)

jonas-schievink:
Confirmed, world gen lag is completely gone, even on my weakest machine! :sparkles: :godmode:

PR #94, opened 2016-07-26
Implement Plantrendering

NilsAffing:
Plants can be rendered
Sticks are rendered correctly

### DIFF
--- a/client/src/view/plant_dummy.vert
+++ b/client/src/view/plant_dummy.vert
@@ -6,18 +6,14 @@ in vec3 normal;
 
 out vec3 x_color;
 out vec3 surfaceNormal;
-out vec3 toLight;
 
 uniform vec3 offset;
 uniform mat4 proj_matrix;
 uniform mat4 view_matrix;
 
-const vec3 sun = vec3(1.0, 0.0, 1.0);
-
 void main() {
     gl_Position = proj_matrix * view_matrix * vec4(position + offset, 1);
 
     surfaceNormal = normal;
-    toLight = sun;
     x_color = color;
 }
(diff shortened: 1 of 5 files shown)


client/src/view/plant_view.rs
```diff
@@ -22,30 +25,30 @@ impl PlantView {
                                  facade: &F)
                                  -> Self {
         // FIXME handle other plant types
-        let mut verts = 0;
+        // let verts = 0;
```
jonas-schievink: Remove this comment

client/src/view/plant_dummy.vert
```diff
@@ -2,13 +2,22 @@
 
 in vec3 position;
 in vec3 color;
+in vec3 normal;
+
 out vec3 x_color;
+out vec3 surfaceNormal;
+out vec3 toLight;
```
LukasKalbertodt: This is currently a vertex attribute but it holds the same value for all vertices. Use a uniform variable instead that you can access in both shaders.

client/src/world/chunk_std.frag
```diff
@@ -7,6 +7,6 @@ in vec3 surfaceNormal;
 out vec3 color;
 
 void main() {
-    float diffuse = max(0.0, dot(toLight, surfaceNormal));
+    float diffuse = max(0.0, dot(normalize(toLight), normalize(surfaceNormal)));
```
jonas-schievink: Is this necessary?

NilsAffing: Yes, because we calculate the normals dynamically and the length of these vectors are not 1.

NilsAffing: Never mind ;)

PR #95, opened 2016-07-26
add weather effects

NiklasPaulRothe:
Work in progress

### DIFF
--- a/client/src/weather/mod.rs
+++ b/client/src/weather/mod.rs
@@ -1,0 +1,248 @@
+extern crate rand;
+
+use glium::{self, DrawParameters, Program, VertexBuffer};
+use glium::draw_parameters::DepthTest;
+use super::camera::Camera;
+use util::ToArr;
+use base::math::*;
+use GameContext;
+use std::rc::Rc;
+
+const BOX_SIZE: f32 = 20.0;
+
+#[derive(Copy, Clone)]
+pub struct Vertex {
+    point: [f32; 3],
+}
+implement_vertex!(Vertex, point);
+
+#[derive(Debug, Copy, Clone)]
+pub enum Form {
+    #[allow(dead_code)]
+    None = 0,
+    #[allow(dead_code)]
+    Rain = 1,
+    #[allow(dead_code)]
+    Snow = 2,
+    #[allow(dead_code)]
+    Pollen = 3,
+}
+
+#[derive(Copy, Clone)]
+pub struct Particle {
+    vertices: [Vertex; 4],
+    position: Point3<f32>,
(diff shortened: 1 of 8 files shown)


client/src/weather/mod.rs
```diff
@@ -0,0 +1,233 @@
...
+pub struct Particle {
+    vertices: [Vertex; 4],
+    position: Point3<f32>,
+    velocity: f32,
+    form: Form,
+}
+
+impl Particle {
+    pub fn new(cam_pos: Point3<f32>, form: Form) -> Particle {
+        let tmp: f32 = rand::random::<f32>() * 6.28;
```
jonas-schievink: Can you generate this number when it's needed instead?

client/src/weather/mod.rs
```diff
@@ -0,0 +1,233 @@
...
+        let mut radius: f32 = 0.0;
+        let mut spawn = false;
+
+        while !spawn {
+            radius = rand::random::<f32>();
+
+            let dice = rand::random::<f32>();
+
+            spawn = dice < radius;
+        }
```
jonas-schievink: This loop is supposed to make larger particles more likely, right? Regardless, can you add a comment explaining what it does?

jonas-schievink: Okay nevermind, `radius` isn't the radius of the particle. Can you comment on that, too? :D (I now have no clue what this loop is doing)

client/src/weather/mod.rs
```diff
@@ -0,0 +1,233 @@
...
+            if len > 20.0 || len < -20.0 {
+                particle.position.x = particle.position.x - (2.0 * tmp.x);
+                particle.position.y = particle.position.y - (2.0 * tmp.y);
+            }
+            if particle.position.z < self.camera.position.z - 20.0 {
+                particle.position.z += 40.0;
+            }
+            if particle.position.z > self.camera.position.z + 20.0 {
+                particle.position.z -= 40.0;
+            }
```
jonas-schievink: It would be nice if you could put the `20.0` in a `const` somewhere above these `if`s, and replace these magic numbers with it.

client/src/weather/mod.rs
```diff
@@ -0,0 +1,233 @@
...
+implement_vertex!(Vertex, position);
+
+#[derive(Debug)]
+pub enum Form {
+    #[allow(dead_code)]
+    Rain,
+    #[allow(dead_code)]
+    Snow,
+    #[allow(dead_code)]
+    Poll,
```
jonas-schievink: `Pollen` is the correct name

PR #96, opened 2016-07-26
helper function to easily load shader files

### LINKED ISSUE #65: Helper functions to easily load shader files
Currently, we use `include_str!()` to include shader files. While this has its advantages, in the next week many groups will work on shaders and might not want to recompile the application everytime a shader is changed. I don't even know if cargo is currently recompiling, which would be even worse because a shader change wouldn't change the compiled program at all...

So yes: a helper function that takes one filename stem, loads "<stem>.frag" and "<stem>.vert" and returns a `Program`.

HelenaKeller:
Closes #65


client/src/game_context.rs
```diff
@@ -22,4 +27,37 @@ impl GameContext {
     pub fn get_config(&self) -> &Config {
         &self.config
     }
+
+    /// Function to load shader files more easily
+    pub fn shader_func<F: Facade>(shader: &str, facade: &F) -> Program {
```
LukasKalbertodt: - `shader_func` is a rather bad name :wink: what about `load_program`?
- The documentation should be extended: What files are loaded? With what extensions? What about shaders other than vertex and fragment? (we should add loading other shaders later)
- I thought about making this a non-static method (taking `&self` as parameter). This would make it possible to do some fun things in the future. I think we should already change it and remove the `facade: &F` parameter, because the facade can be obtained through `&self`

client/src/game_context.rs
```diff
@@ -22,4 +27,37 @@ impl GameContext {
...
+        if (Path::new(format!("shader/{}.vert", shader).as_str())).exists() &&
+           (Path::new(format!("shader/{}.frag", shader).as_str())).exists() {
```
LukasKalbertodt: How creating the two `Path` objects before and saving them as `vert_path` and `frag_path`? This saves some duplicated code.

client/src/game_context.rs
```diff
@@ -22,4 +27,37 @@ impl GameContext {
...
+
+            let mut vert = match File::open((format!("shader/{}.vert", shader)).as_str()) {
+                Err(_) => panic!("Could not open {}.vert!", shader),
+                Ok(file) => file,
+            };
+            let mut frag = match File::open((format!("shader/{}.frag", shader)).as_str()) {
+                Err(_) => panic!("Could not open {}.frag!", shader),
+                Ok(file) => file,
+            };
```
LukasKalbertodt: If we return `Box<Error>`, we can replace both matches by

```rust
let vert = try!(File::open(vert_path));
let frag = try!(File::open(frag_path));
```

client/src/game_context.rs
```diff
@@ -22,4 +27,37 @@ impl GameContext {
...
+            let mut frag_buf = String::new();
+
+            match vert.read_to_string(&mut vert_buf) {
+                Err(_) => panic!("Could not read {}.vert!", shader),
+                Ok(_) => (),
+            }
+            match frag.read_to_string(&mut frag_buf) {
+                Err(_) => panic!("Could not read the {}.frag!", shader),
+                Ok(_) => (),
+            }
```
LukasKalbertodt: Same as above: replace `match` by `try!`

client/src/game_context.rs
```diff
@@ -22,4 +27,37 @@ impl GameContext {
...
+
+            Program::from_source(facade, vert_buf.as_str(), frag_buf.as_str(), None).unwrap()
```
LukasKalbertodt: We should remove the `unwrap()` and wrap the remaining line into `try!()`. You should also be able to just write `&vert_buf` instead of `vert_buf.as_str()`.

client/src/game_context.rs
```diff
@@ -22,4 +27,37 @@ impl GameContext {
...
+        } else {
+            panic!("No such shader file!")
+        }
```
LukasKalbertodt: We can remove the whole if-else, when using `Box<Error>` and `try!(File::open())`.

client/src/view/plant_renderer.rs
```diff
@@ -1,19 +1,19 @@
...
-        let prog = Program::from_source(facade,
-                                        include_str!("plant_dummy.vert"),
-                                        include_str!("plant_dummy.frag"),
-                                        None)
-            .unwrap();
-
-        PlantRenderer { program: prog }
+    pub fn new(context: Rc<GameContext>) -> Self {
+        PlantRenderer {
+            program: match context.load_program("plant_dummy") {
```
jonas-schievink: Why not `.unwrap()`? (or `.expect()`)

jonas-schievink:
If you're going to panic anyways if a shader fails to load, why not just do `.expect("failed to load shader")` instead of matching the returned result?

PR #97, opened 2016-07-26
Add player physics

### LINKED ISSUE #37: Player "physics"
Tomorrow we will probably land a PR that adds a system to handle events properly. On top of that we can start to implement a `Player` that implements the `EventHandler` trait. In contrast to the `Ghost` control, the player is bound to the earth by the laws of physics :stuck_out_tongue: 

To get started, the players x and y coordinate could be controlled by controls while the z coordinate is just set to the world-height at the players position + 1 (or something like that). 

After that some kind of gravity can be implemented and the player could gain the ability to jump (and maybe to duck). 

Complex physical behavior like collision with plants is not in the scope of this issue, though.

JonasKnerr:
closes #37

### DIFF
--- a/base/src/gen/world/mod.rs
+++ b/base/src/gen/world/mod.rs
@@ -193,6 +193,7 @@ impl ChunkProvider for WorldGenerator {
         }))
     }
 
+
     fn is_chunk_loadable(&self, _: ChunkIndex) -> bool {
         // All chunks can be generated from nothing
         true
--- a/base/src/world/world.rs
+++ b/base/src/world/world.rs
@@ -47,7 +47,6 @@ impl World {
         let out = self.chunks.get(&ChunkIndex(chunk_pos)).map(|chunk| {
 
             let mut inner_pos = pos.0 % chunk_size;
-            println!("Inner Pos ------------{:?}", inner_pos);
             if inner_pos.q < 0 {
                 inner_pos.q += chunk_size;
             }
--- a/base/src/world/world.rs
+++ b/base/src/world/world.rs
@@ -5,7 +5,7 @@ use math::*;
 /// Represents a whole game world consisting of multiple `Chunk`s.
 ///
 /// Chunks are parallelograms (roughly) that are placed next to each other
-/// in the world.
+/// in the world.]
 pub struct World {
     // TODO: make it private after we can access it immutable via a method! (see #7)
     pub chunks: HashMap<ChunkIndex, Chunk>,
@@ -40,13 +40,13 @@ impl World {
     /// corresponding chunk is loaded.
     pub fn pillar_at(&self, pos: PillarIndex) -> Option<&HexPillar> {
         let chunk_size = super::CHUNK_SIZE as i32;
+
         // TODO: use `/` operator once it's implemented
         // let chunk_pos = pos / (super::CHUNK_SIZE as i32);
         let chunk_pos = pos.0 / chunk_size;
 
+
         let out = self.chunks.get(&ChunkIndex(chunk_pos)).map(|chunk| {
-            // TODO: use `%` operator once it's implemented
-            // let inner_pos = pos % (super::CHUNK_SIZE as i32);
             let mut inner_pos = pos.0 % chunk_size;
             if inner_pos.q < 0 {
                 inner_pos.q += chunk_size;
@@ -59,15 +59,37 @@ impl World {
             }
             &chunk[inner_pos]
         });
-
+        // println!("out {:?}", out.get(AxialPoint { q: 5.0, r: 7.0 }));
         if out.is_none() {
+            // println!("test");
             debug!("chunk {:?} is not loaded (position request {:?})",
                    chunk_pos,
                    pos);
         }
 
+        // println!("out {:?}", out);
         out
     }
+
+    // pub fn pillar_at_axp(&self, pos: AxialPoint) -> Option<&HexPillar> {
+    //     let out = self.chunks.get(&ChunkIndex(pos)).map(|chunk| {
+    //         // TODO: use `%` operator once it's implemented
+    //         // let inner_pos = pos % (super::CHUNK_SIZE as i32);
+    //         let inner_pos = AxialPoint::new(pos.q % (super::CHUNK_SIZE as i32),
+    //                                         pos.r % (super::CHUNK_SIZE as i32));
+    //         println!("inner pos {:?}", chunk[inner_pos]);
+    //         &chunk[inner_pos]
+    //     });
+
+    //     if out.is_none() {
+    //         println!("test");
+    //         debug!("chunk {:?} is not loaded (position request {:?})", pos, pos);
+    //     }
+
+    //     println!("out {:?}", out);
+    //     out
+    // }
+
 
     /// Returns the chunk in which the given pillar exists.
     pub fn chunk_from_pillar(&self, pos: PillarIndex) -> Option<&Chunk> {
--- a/base/src/world/world.rs
+++ b/base/src/world/world.rs
@@ -5,7 +5,7 @@ use math::*;
 /// Represents a whole game world consisting of multiple `Chunk`s.
 ///
 /// Chunks are parallelograms (roughly) that are placed next to each other
-/// in the world.]
+/// in the world.
 pub struct World {
     // TODO: make it private after we can access it immutable via a method! (see #7)
     pub chunks: HashMap<ChunkIndex, Chunk>,
@@ -35,7 +35,8 @@ impl World {
             Ok(())
         }
     }
-    /// Returns the hex pillar at the given world position, if the
+
+    /// Returns the hex pillar at the given world position, iff the
     /// corresponding chunk is loaded.
     pub fn pillar_at(&self, pos: PillarIndex) -> Option<&HexPillar> {
         let chunk_size = super::CHUNK_SIZE as i32;
@@ -57,6 +58,8 @@ impl World {
             }
             &chunk[inner_pos]
         });
+
+
         if out.is_none() {
             debug!("chunk {:?} is not loaded (position request {:?})",
                    chunk_pos,
--- a/base/src/world/world.rs
+++ b/base/src/world/world.rs
@@ -47,6 +47,7 @@ impl World {
         let out = self.chunks.get(&ChunkIndex(chunk_pos)).map(|chunk| {
 
             let mut inner_pos = pos.0 % chunk_size;
+            println!("Inner Pos ------------{:?}", inner_pos);
             if inner_pos.q < 0 {
                 inner_pos.q += chunk_size;
             }
--- a/base/src/world/world.rs
+++ b/base/src/world/world.rs
@@ -40,10 +40,16 @@ impl World {
     /// corresponding chunk is loaded.
     pub fn pillar_at(&self, pos: PillarIndex) -> Option<&HexPillar> {
         let chunk_size = super::CHUNK_SIZE as i32;
-
+        let mut new_pos = pos;
         // TODO: use `/` operator once it's implemented
         // let chunk_pos = pos / (super::CHUNK_SIZE as i32);
-        let chunk_pos = pos.0 / chunk_size;
+        if new_pos.0.q < 0 {
+            new_pos.0.q -= 15;
+        }
+        if new_pos.0.r < 0 {
+            new_pos.0.r -= 15;
+        }
+        let chunk_pos = new_pos.0 / chunk_size;
         let out = self.chunks.get(&ChunkIndex(chunk_pos)).map(|chunk| {
 
             let mut inner_pos = pos.0 % chunk_size;
--- a/base/src/world/world.rs
+++ b/base/src/world/world.rs
@@ -35,8 +35,7 @@ impl World {
             Ok(())
         }
     }
-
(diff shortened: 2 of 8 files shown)


base/src/gen/world/mod.rs
```diff
@@ -56,9 +56,13 @@ impl ChunkProvider for WorldGenerator {
...
         }
 
+
+
+
         Some(Chunk::from_pillars(pillars))
     }
 
+
     fn is_chunk_loadable(&self, _: ChunkIndex) -> bool {
```
LukasKalbertodt: Random newlines? :wink:

LukasKalbertodt:
You checked in quite a few `tmp****.rs` files. Please remove those and avoid using `git add .` or similar "add everything" commands in the future.

base/src/world/world.rs
```diff
@@ -5,7 +5,7 @@ use math::*;
 /// Represents a whole game world consisting of multiple `Chunk`s.
 ///
 /// Chunks are parallelograms (roughly) that are placed next to each other
-/// in the world.
+/// in the world.]
```
LukasKalbertodt: I guess this is by accident?

base/src/world/world.rs
```diff
@@ -35,29 +35,34 @@ impl World {
             Ok(())
         }
     }
-
-    /// Returns the hex pillar at the given world position, iff the
+    /// Returns the hex pillar at the given world position, if the
```
LukasKalbertodt: The `iff` was intended. It's a short form of "if and only if". Please re-add the removed empty line, too.

base/src/world/world.rs
```diff
@@ -35,29 +35,34 @@ impl World {
...
-                                        pos.0.r / (super::CHUNK_SIZE as i32));
+        let chunk_size = super::CHUNK_SIZE as i32;
 
+        let chunk_pos = pos.0 / chunk_size;
         let out = self.chunks.get(&ChunkIndex(chunk_pos)).map(|chunk| {
-            // TODO: use `%` operator once it's implemented
-            // let inner_pos = pos % (super::CHUNK_SIZE as i32);
-            let inner_pos = AxialPoint::new(pos.0.q % (super::CHUNK_SIZE as i32),
-                                            pos.0.r % (super::CHUNK_SIZE as i32));
+
```
LukasKalbertodt: This empty line can be removed (avoid empty lines after the closing `{`)

base/src/world/world.rs
```diff
@@ -35,29 +35,34 @@ impl World {
...
+            }
+            if inner_pos.r < 0 {
+                inner_pos.r += chunk_size;
+            }
+            // mistery how this works but it works
+            if chunk_pos.q != chunk_pos.r {
+                let tmp = inner_pos.q;
+                inner_pos.q = inner_pos.r;
+                inner_pos.r = tmp;
+            }
```
LukasKalbertodt: Typo: "mystery". Please also refer to the issue with `(see #98)`.

You can easily swap with `::std::mem::swap(&mut inner_pos.r, &mut inner_pos.q)`

base/src/world/world.rs
```diff
@@ -35,29 +35,34 @@ impl World {
...
             &chunk[inner_pos]
         });
-
         if out.is_none() {
             debug!("chunk {:?} is not loaded (position request {:?})",
                    chunk_pos,
                    pos);
         }
-
```
LukasKalbertodt: Those two empty lines could have stayed there :stuck_out_tongue:

client/src/world/chunk_view.rs
```diff
@@ -56,6 +56,9 @@ impl ChunkView {
 
         let mut pillars = Vec::new();
         for q in 0..world::CHUNK_SIZE * world::CHUNK_SIZE {
+            let v = AxialVector::new((q / world::CHUNK_SIZE).into(),
+                                     (q % world::CHUNK_SIZE).into());
+
```
LukasKalbertodt: I think I added this for debugging. So this can be removed (Rust should warn about this)

client/src/player.rs
```diff
@@ -0,0 +1,268 @@
...
+const VELOCITY: f32 = 1.5;
+const LOW_VELOCITY: f32 = 1.0;
+const FAST_VELOCITY: f32 = 10.0;
+pub struct Player {
+    cam: Camera,
+    context: Rc<GameContext>,
+    world_manager: WorldManager,
+    walk_vel: Vector2<f32>,
+    dz: f32,
+    gravity: f32,
```
LukasKalbertodt: This can be a constant, too, as it's never changed.

client/src/player.rs
```diff
@@ -0,0 +1,268 @@
...
+
+const VELOCITY: f32 = 1.5;
+const LOW_VELOCITY: f32 = 1.0;
+const FAST_VELOCITY: f32 = 10.0;
+pub struct Player {
+    cam: Camera,
+    context: Rc<GameContext>,
+    world_manager: WorldManager,
+    walk_vel: Vector2<f32>,
+    dz: f32,
```
LukasKalbertodt: Prefer more descriptive names... I don't have a clue what `dz` could be... 

And if it's the velocity on the z-axis, consider combining it with `walk_vel` and call it `velocity: Vector3f`

client/src/player.rs
```diff
@@ -0,0 +1,268 @@
...
+use super::world_manager::*;
+
+const VELOCITY: f32 = 1.5;
+const LOW_VELOCITY: f32 = 1.0;
+const FAST_VELOCITY: f32 = 10.0;
+pub struct Player {
+    cam: Camera,
+    context: Rc<GameContext>,
+    world_manager: WorldManager,
+    walk_vel: Vector2<f32>,
```
LukasKalbertodt: We have nice little typedefs: you can say `Vector2f` to refer to a `Vector2<f32>`

client/src/player.rs
```diff
@@ -0,0 +1,268 @@
...
+            up: false,
+            down: false,
+            mouselock: false,
+            speed: false,
+        }
+    }
+
+    pub fn get_height(&self) -> f32 {
+        let world = self.world_manager.get_world();
+        let real_pos = Point2f::new(self.cam.position[0], self.cam.position[1]);
```
LukasKalbertodt: Consider using `position.x` and `position.y` instead: it's statically safer and more expressive.

client/src/player.rs
```diff
@@ -0,0 +1,268 @@
...
+        let axial_pos = AxialPoint::from_real(real_pos);
+        let argument = PillarIndex(axial_pos);
```
LukasKalbertodt: Maybe combine those lines to `let pillar_index = PillarIndex(AxialPoint::from_real(real_pos));`

client/src/player.rs
```diff
@@ -0,0 +1,268 @@
...
+            left: false,
+            right: false,
+            up: false,
+            down: false,
+            mouselock: false,
+            speed: false,
+        }
+    }
+
+    pub fn get_height(&self) -> f32 {
```
LukasKalbertodt: How about call it `get_ground_height` and return an `Option<f32>` instead.

client/src/player.rs
```diff
@@ -0,0 +1,268 @@
...
+        let axial_pos = AxialPoint::from_real(real_pos);
+        let argument = PillarIndex(axial_pos);
+        let pillar = world.pillar_at(argument);
+
+        if pillar.is_some() {
+            let sect = pillar.unwrap().sections();
+            sect[0].top.to_real()
+        } else {
+            0.0
+        }
```
LukasKalbertodt: When returning an `Option` we don't need to return a stinky default value of `0.0` but can just say: 

```rust
world.pillar_at(pillar_index).map(|pillar| p.sections()[0].top.to_real())
```

client/src/player.rs
```diff
@@ -0,0 +1,268 @@
...
+    pub fn get_camera(&self) -> Camera {
+        self.cam
+    }
+
+    pub fn jump(&mut self) {
+
+        if self.is_jumping == false {
+            self.dz = 0.11;
+            self.is_jumping = true;
+        }
```
LukasKalbertodt: Shouldn't we check `is_falling`, too? (also: remove the empty line in the function)

client/src/player.rs
```diff
@@ -0,0 +1,268 @@
...
+    }
+
+    pub fn update(&mut self, delta: f32) {
+        let height = self.get_height() + 1.0;
```
LukasKalbertodt: Please explain magic constants (like this `1.0`) with a comment

client/src/player.rs
```diff
@@ -0,0 +1,268 @@
...
+
+        if !self.is_jumping && !self.is_falling {
+            if height < self.cam.position[2] {
```
LukasKalbertodt: Use `position.z` instead (see above for explanation)

client/src/player.rs
```diff
@@ -0,0 +1,268 @@
...
+                self.is_falling = true;
+            } else {
+                self.cam.position[2] = height;
+            }
+        }
+
+        if self.cam.position[2] > height && !self.is_jumping {
```
LukasKalbertodt: `.z` again

client/src/player.rs
```diff
@@ -0,0 +1,268 @@
...
+            self.is_falling = true;
+        }
+
+        if self.is_jumping {
+            self.is_falling = true;
+        }
```
LukasKalbertodt: I'm pretty sure you can and *should* combine `is_jumping` and `is_falling` into one variable (e.g. `is_falling`). I don't think we really need to distinguish between the two states.

client/src/player.rs
```diff
@@ -0,0 +1,268 @@
...
+        if self.is_falling {
+            self.dz = self.dz - self.gravity;
```
LukasKalbertodt: You should use `delta` here, as gravity is the acceleration in m/s²

client/src/player.rs
```diff
@@ -0,0 +1,268 @@
...
+
+            if self.dz < -0.2 {
+                self.dz = -0.2;
+            }
```
LukasKalbertodt: This is a hack to simulate [terminal velocity](https://en.wikipedia.org/wiki/Terminal_velocity), which is around 54m/s for a human (skydiver in belly-to-earth position). I think this is fine for now, but we should replace it with a proper drag calculation later.

client/src/player.rs
```diff
@@ -0,0 +1,268 @@
...
+            self.cam.move_up(self.dz);
```
LukasKalbertodt: You need to use `delta` here.

LukasKalbertodt:
I made a bunch of line comments. I hope you don't mind, but I think you can save a lot of duplicate code and use a much more elegant design by rewriting a lot of your code.

First of all: you should use the metric system for everything. This means: meter for distances (or positions), m/s for velocities and m/s² for accelerations. 

It might be interesting and nice to create a type system for this. Such that your velocity is of type `MeterPerSeconds` instead of `f32`. Something similar exists in [the `dimensioned` crate](https://github.com/paholg/dimensioned), which is unstable for the moment (we can't use it). This would catch a bunch of bugs at compile time (such as forgetting to multiply by `delta`, which you forgot a lot!). But I think the type system thingy is overkill for now. Interested in your opinion anyway, @jonas-schievink 

Next: only use one easy vector: `velocity: Vector3f`. We can then remove nearly all boolean flags (except the ones for the mouse) by adding some value onto `velocity` directly in `handle_event()` whenever a button is pressed. Ok, I lied: sadly it's not *that* easy, because we can't access `delta` in `handle_event()`. So we add a third vector `acceleration: Vector3f`. We "collect" all force acting upon our player in that vector (not only movement by keys but gravity, too). After that we apply those forces in the `update()` method like so:

```rust
self.velocity += self.acceleration * delta
```

Note that this calculation conforms with the units: `m/s = (m/s²) * s`.

Next, we need to apply friction. Calculating correct friction sounds crazy, so we can just simulate it by subtracting a part of the current velocity from the current velocity. Something like:

```rust
self.velocity -= self.velocity.mul_element_wise(Vector3::new(0.3, 0.3, 0.02));
```

Note that I used different factors for x/y and z, because x/y describes the friction from the ground while z describes the friction from air (which is much lower). However, this is of course very hacky and physically incorrect.

Lastly, we just need to change the position:

```rust
self.camera.position += self.velocity * delta;   // units work perfectly again
``` 

Also try to combine `is_jumping` and `is_falling` into one variable. I'm pretty sure, it's possible. Your code can be reduced in size, which is a good thing here. If anything is unclear, just ask me tomorrow! :simple_smile:

jonas-schievink:
Yeah, explicit types for these things are neat but make code harder to understand, let's not do it right now.

LukasKalbertodt:
I don't have time to review this, but: we can't merge it like that even when the code looks nice. Other groups need to use the ghost player to fly around and look at their changes. 

So your next task would be to add another type (maybe `ControlSwitcher`) that owns the ghost and player and can switch between those. This means that `ControlSwitcher` also has a `get_camera()` method that returns the camera of the active control method. And `ControlSwitcher` implements `EventReceiver` and forwards the events accordingly *and* listens on the `[g]` button that allows the player to switch between player and ghost. Something like that.

LukasKalbertodt:
You added a `tmp***.rs` again...

I think I won't really try to understand this PR, but simply merge if it works and after it has been rebased.

PR #103, opened 2016-07-27
add Biomes, Ground Materials, Plant placement

adRichter:


base/src/gen/world/mod.rs
```diff
@@ -22,25 +27,81 @@ pub struct WorldGenerator {
...
     /// Returns the seed of this world generator.
     pub fn seed(&self) -> u64 {
         self.seed
     }
+
+    fn biome_from_climate(temperature: f32, humidity: f32) -> Biome {
+        match (temperature, humidity) {
+            (0.0...0.2, 0.0...0.2) => Biome::Stone,
+            (0.0...0.2, 0.2...0.4) => Biome::Snow,
+            (0.0...0.2, 0.4...1.0) => Biome::Snow,
```
LukasKalbertodt: Those two arms could be replaced by one `(0.0...0.2, 0.0...1.0) => Biome::Snow`, right?

base/src/gen/world/mod.rs
```diff
@@ -22,25 +27,81 @@ pub struct WorldGenerator {
...
+            (0.2...0.4, 0.0...0.2) => Biome::GrassLand,
+            (0.2...0.4, 0.2...0.4) => Biome::GrassLand,
```
LukasKalbertodt: Same here

base/src/gen/world/mod.rs
```diff
@@ -51,77 +112,97 @@ impl ChunkProvider for WorldGenerator {
...
+
             let real_pos = pos.to_real();
             let x = real_pos.x;
             let y = real_pos.y;
-
             // Pillar pos relative to first pillar
-            let rel_pos = pos - index.0 * CHUNK_SIZE as i32;
-            let column = &fill[rel_pos.q as usize][rel_pos.r as usize];
+            let rel_q = pos.q - index.0.q * CHUNK_SIZE as i32;
+            let rel_r = pos.r - index.0.r * CHUNK_SIZE as i32;
```
LukasKalbertodt: Those `AxialVectors` implement the basic operations for a reason :wink: `let rel = pos - index * (CHUNK_SIZE as i32)`

base/src/gen/world/mod.rs
adRichter: Yeah, in this case for sure, but those are generally nice borders. So this way you can easily change biomes later on.

base/src/gen/world/mod.rs
adRichter: Same here ;)

base/src/gen/world/mod.rs
```diff
@@ -1,17 +1,22 @@
...
-
+enum Biome {
+    GrassLand,
+    Desert,
+    Snow,
+    Forest,
+    RainForest,
+    Savanna,
+    Stone,
+    WeihnachtsmannLand,
```
jonas-schievink: Change this to an english name? (or keep it as a joke?)

base/src/gen/world/mod.rs
```diff
@@ -22,25 +27,81 @@ pub struct WorldGenerator {
...
+    }
+
+    /// trying aproximate the
+    /// steepvalues       10   20   60  200
+    /// for temperature  0.0  0.2  0.4  1.0
+    fn steepness_from_temperature(temperature: f32) -> f32 {
+        120.0 * temperature * temperature + 72.0 * temperature + 3.5
+    }
+
+    fn material_from_biome(biome: &Biome) -> GroundMaterial {
```
jonas-schievink: This method makes more sense in an `impl Biome`, so you could just do `my_biome.material()`.

base/src/gen/world/mod.rs
```diff
@@ -22,25 +27,81 @@ pub struct WorldGenerator {
...
+        match *biome {
+            Biome::GrassLand => GroundMaterial::Grass,
+            Biome::Desert => GroundMaterial::Sand,
+            Biome::Snow => GroundMaterial::Snow,
+            Biome::Forest => GroundMaterial::Mulch,
+            Biome::RainForest => GroundMaterial::Jungle,
+            Biome::Savanna => GroundMaterial::Dirt,
+            Biome::Stone => GroundMaterial::Stone,
+            Biome::WeihnachtsmannLand => GroundMaterial::Color(1.0, 0.0, 0.0),
```
jonas-schievink: Shouldn't this be `Snow` instead of red color? Or remove `Biome::WeihnachtsmannLand` in favor of just `Biome::Snow`?

base/src/world/ground.rs
```diff
@@ -3,6 +3,12 @@ pub enum GroundMaterial {
     Dirt,
     Grass,
     Stone,
+    Sand,
+    Snow,
+    Jungle,
+    Mulch,
+    Grey(f32),
+    Color(f32, f32, f32),
```
jonas-schievink: `Grey` can be removed, you can do the same thing with `Color`. I think `Color` should be removed as well, since it doesn't have any meaningful material properties.

base/src/world/ground.rs
```diff
@@ -3,6 +3,12 @@ pub enum GroundMaterial {
     Dirt,
     Grass,
     Stone,
+    Sand,
+    Snow,
+    Jungle,
```
jonas-schievink: `JungleGrass` might be a better name ("Jungle" isn't really any kind of material)

base/src/gen/world/mod.rs
KarolinePlum: we needed a default biome so that the compiler keeps quiet. this biome will in theory never appear but if it does we think it would be nice for it to immediately catch the eye so that we can fix it.

PR #106, opened 2016-07-27
Time

FlorianJanosch:
add daytime.rs with Struct DayTime
DayTime has:
`time_year` which counts the years
`time_day` which counts the days
`time_on_day` which counts the time of the current day

implement EventHandler for DayTime
add `Add` (Plus-Key) to speed up time by factor 1_000.0, for demonstration purposes in the future

Game now calls update function of daytime with paramter delta in main-loop
put daytime in the list of EventHandlers, that the event_manager iterates through

### DIFF
--- a/client/src/daytime.rs
+++ b/client/src/daytime.rs
@@ -1,0 +1,93 @@
+use super::event_manager::*;
+use glium::glutin::{ElementState, Event, VirtualKeyCode};
+
+#[derive(Debug)]
+pub struct DayTime {
+    time_year: u32,
+    time_day: u32,
+    time_on_day: f32,
+    speed: f32,
+}
+
+// `DEFAULT_TIME_SPEED` is 1.0 so the time goes at normal speed
+// `PLUS_TIME_SPEED` is the factor with which the time is sped up, when the
+// speed-up key is pressed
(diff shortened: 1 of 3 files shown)


client/src/daytime.rs
```diff
@@ -0,0 +1,134 @@
...
+const PLUS_TIME_SPEED: f32 = 1000.0;
+
+
+impl Default for DayTime {
+    fn default() -> DayTime {
+        DayTime {
+            time_year: 0,
+            time_day: 0,
+            time_on_day: 0.0,
+            speed: 1.0,
```
jonas-schievink: Use `DEFAULT_TIME_SPEED` instead of `1.0` to be more explicit

client/src/daytime.rs
```diff
@@ -0,0 +1,134 @@
...
+        }
+
+    }
+}
+
+
+
+// position in theta phi
+
+// lichtvektor
```
jonas-schievink: Unnecessary comments?

client/src/lib.rs
```diff
@@ -20,6 +20,9 @@ mod event_manager;
 pub mod view;
 pub mod util;
 
+pub mod daytime;
```
jonas-schievink: The empty line above can go away

PR #112, opened 2016-07-28
dem SUN

jbesteuos:


client/src/view/sun_view.rs
```diff
@@ -0,0 +1,120 @@
...
+        let vbuf = VertexBuffer::new(context.get_facade(), &raw_vertex_buffer).unwrap();
+
+        let raw_index_buffer = [2, 1, 0, 3]; //TrianglesStrip
+
+        let ibuf = IndexBuffer::new(context.get_facade(),
+                                    PrimitiveType::TriangleStrip,
+                                    &raw_index_buffer)
+            .unwrap();
+
+        let prog = Program::from_source(context.get_facade(),
```
jonas-schievink: You can use `GameContext::load_program` instead (and move the shaders to the shader folder). This way you won't have to recompile plantex when changing a shader

client/src/view/sun_view.rs
```diff
@@ -0,0 +1,120 @@
...
+                  &params)
+            .unwrap();
+    }
+
+    pub fn update(&mut self, pos: Vector3f) {
+
+        const SUN_SIZE: f32 = 35.0;
+        self.position = pos;
+        let raw_vertex_buffer = vec![
+            Vertex { i_position: [self.position.x+SUN_SIZE,
```
jonas-schievink: I don't think this is formatted right

client/src/view/sun_view.rs
```diff
@@ -0,0 +1,120 @@
...
+
+        self.vertex_buffer = VertexBuffer::new(self.context.get_facade(), &raw_vertex_buffer)
+            .unwrap();
+
+    }
+}
+#[derive(Debug, Copy, Clone)]
+struct Vertex {
+    pub i_position: [f32; 3],
+    pub i_unit_coords: [f32; 3],
```
jonas-schievink: The third coordinate is always 0, could you use a `[f32; 2]` instead?

jbesteuos: No, third component is for the height

PR #113, opened 2016-07-28
Bloom

gustav1101:
Currently still WIP, but bloom is already fully implemented and working. Currently refactoring code to align it with project standards, making small changes to make code more efficient & easier to read, and implementing possibility to really switch off bloom.


client/src/renderer.rs
```diff
@@ -75,23 +111,161 @@ impl Renderer {
...
+                bloom_blur_horz_buffer.draw(&self.quad_vertex_buffer,
+                          &self.quad_index_buffer,
+                          &self.context.load_program("bloom_blur").unwrap(),
+                          &uniforms_horz_blur,
+                          &Default::default())
+                    .unwrap();
+            } else {
+                bloom_blur_vert_buffer.draw(&self.quad_vertex_buffer,
+                          &self.quad_index_buffer,
+                          &self.context.load_program("bloom_blur").unwrap(),
```
jonas-schievink: You should preload the program and put it in the renderer (or store it in some other struct), this is just horribly slow :D

client/src/renderer.rs
```diff
@@ -75,23 +111,161 @@ impl Renderer {
...
+        // &Default::default())
+        // .unwrap();
+        //
+        // horizontal = false;
+
+        for _ in 0..10 {
+            if horizontal {
+                bloom_blur_horz_buffer.draw(&self.quad_vertex_buffer,
+                          &self.quad_index_buffer,
+                          &self.context.load_program("bloom_blur").unwrap(),
```
jonas-schievink: (see other comment)

client/src/renderer.rs
```diff
@@ -75,23 +111,161 @@ impl Renderer {
...
+        };
+
+
+        let mut target = self.context.get_facade().draw();
+        target.clear_color_and_depth((0.0, 0.0, 0.0, 1.0), 1.0);
+
         target.draw(&self.quad_vertex_buffer,
                   &self.quad_index_buffer,
-                  &self.tonemapping_program,
+                  &self.context.load_program("tonemapping").unwrap(),
```
jonas-schievink: Same here. You already had the tonemapping program in a field.

client/shader/bloom.vert
```diff
@@ -0,0 +1,13 @@
...
+layout(location = 1) in vec2 in_texcoord;
+
+out VertexData {
+    vec2 frag_texcoord;
+} o;
+
+void main() {
+    o.frag_texcoord = in_texcoord;
+    gl_Position = in_position;
+}
```
krschmidt1: bloom.vert, bloom_blending.vert, bloom_blur.vert and tonemapping.vert are all exactly the same. You could instead have only one vertex shader file and use it for several programs. (maybe call it something like screenquad.vert)

jonas-schievink: That only works if the fragment shader has the same name, since the loader looks for `{name}.vert` and `{name}.frag`. So if the fragment shaders are *also* the same it could work.

client/shader/bloom_blending.frag
```diff
@@ -0,0 +1,17 @@
...
+} i;
+
+layout(location = 0) out vec4 out_color;
+
+void main()
+{
+    vec3 hdr_color = texture(world_tex, i.frag_texcoord).rgb;
+    vec3 bloom_color = texture(bloom_tex, i.frag_texcoord).rgb;
+    out_color = vec4(hdr_color + bloom_color, 1.0);
+}
```
krschmidt1: If you don't plan to make this much more complex, this could also be done in the final composition fragment shader (atm: tonemapping.frag). This would save one rendering of the screenquad + some texture read/write

gustav1101:
Travis doesn't complete, closing this and reopening

PR #117, opened 2016-07-28
Bloom

gustav1101:
Clone of #113 

will add possibility to completly disable bloom in the future, but would like to merge this version


client/shader/bloom_filter.frag
```diff
@@ -0,0 +1,21 @@
...
+} i;
+
+layout(location = 0) out vec4 out_color;
+
+void main() {
+    vec3 col = texture(decal_texture, i.frag_texcoord).rgb;
+
+
+    // transform to greyscale for proper brightness
+    if (dot(col, vec3(0.2126, 0.7152, 0.0722)) > 0.5) {
```
jonas-schievink: Where do these values come from? This looks like RGB->YUV conversion on first glance, but the values don't seem to be correct for that. Please add a comment explaining what exactly this does (and maybe add a link).

client/shader/bloom_blur.frag
```diff
@@ -0,0 +1,37 @@
...
+out vec4 FragColor;
+//in vec2 frag_texcoord;
+in VertexData {
+    vec2 frag_texcoord;
+} i;
+uniform sampler2D image;
+
+uniform bool horizontal;
+
+uniform float weight[5] = float[] (0.227027, 0.1945946, 0.1216216, 0.054054, 0.016216);
```
jonas-schievink: Can you add a comment explaining what this shader does exactly? Is this gaussian blur or something else?

client/shader/bloom_blur.frag
```diff
@@ -0,0 +1,37 @@
+#version 330 core
+out vec4 FragColor;
+//in vec2 frag_texcoord;
```
jonas-schievink: Remove this

client/shader/bloom_blending.frag
```diff
@@ -0,0 +1,17 @@
...
+
+in VertexData {
+    vec2 frag_texcoord;
+} i;
+
+layout(location = 0) out vec4 out_color;
+
+void main()
+{
+    vec3 hdr_color = texture(world_tex, i.frag_texcoord).rgb;
```
jonas-schievink: Does bloom care about whether HDR is actually used? If not, `hdr_color` should be named something else, like `world_color`

client/shader/bloom_blur.frag
```diff
@@ -0,0 +1,37 @@
...
+    else
+    {
+        for(int i = 1; i < 5; ++i)
+        {
+            result += texture(image, TexCoords + vec2(0.0, tex_offset.y * i)).rgb * weight[i];
+            result += texture(image, TexCoords - vec2(0.0, tex_offset.y * i)).rgb * weight[i];
+        }
+    }
+    FragColor = vec4(result, 1.0);
+    //FragColor = vec4(TexCoords, 0.0, 1.0);
```
jonas-schievink: Remove commented code here too

client/src/renderer.rs
```diff
@@ -42,19 +50,60 @@ impl Renderer {
...
+                                                          UncompressedFloatFormat::F32F32F32F32,
+                                                          MipmapsOption::NoMipmap,
+                                                          resolution.0,
+                                                          resolution.1)
+            .unwrap();
+
+
+
+        let bloom_blend_tex = Texture2d::empty_with_format(context.get_facade(),
+                                                           UncompressedFloatFormat::F32F32F32F32,
```
jonas-schievink: Do we actually need an alpha channel everywhere? You seem to use it in `bloom_filter.frag` (you assign different values to A), but the next stage, `bloom_blur.frag`, seems to completely ignore it.

client/src/renderer.rs
```diff
@@ -42,19 +50,60 @@ impl Renderer {
...
-        let tonemapping_program = Program::from_source(context.get_facade(),
-                                                       include_str!("tonemapping.vert"),
-                                                       include_str!("tonemapping.frag"),
-                                                       None)
+
+        let bloom_quad_tex = Texture2d::empty_with_format(context.get_facade(),
+                                                          UncompressedFloatFormat::F32F32F32F32,
+                                                          MipmapsOption::NoMipmap,
+                                                          resolution.0,
+                                                          resolution.1)
```
jonas-schievink: We now have a `render_update` method that is called whenever the screen size changes, so this should go in there (maybe we need to rebase first).

client/shader/bloom_blending.frag
gustav1101: Yes, it does. I mean, you could run bloom without HDR, but that would just result in strange effects. Bloom really makes most sense with HDR, and since we can rely on the incoming values actually being HDR (unless someone were to exchange the HDR-buffer and all hdr-textures used in the game) I think this is justified. If you disagree tell me and I'll change it.

client/src/renderer.rs
gustav1101: Tried to change this, but it caused problems and will need time to fix. Will do this when we do optimization and have time.

PR #118, opened 2016-07-28
advanced weather effects

NiklasPaulRothe:


client/src/weather/mod.rs
```diff
@@ -28,13 +29,24 @@ pub enum Form {
     Pollen = 3,
 }
 
+#[derive(Debug, Copy, Clone)]
+pub enum Strength {
+    #[allow(dead_code)]
+    Weak = 1,
+    #[allow(dead_code)]
+    Medium = 2,
+    #[allow(dead_code)]
```
LukasKalbertodt: Are those `dead_code` annotations necessary? Since the enum is public, it shouldn't warn about it.

client/src/weather/mod.rs
```diff
@@ -81,6 +93,7 @@ impl Particle {
             velocity: velocity,
             trans_x: rand::random::<f32>() * 6.28,
             trans_y: rand::random::<f32>() * 6.28,
+            trans_z: rand::random::<f32>() * 6.28,
```
LukasKalbertodt: Use `std::f32::consts::PI * 2` instead... this is a very rough approximation 0_o

client/src/weather/mod.rs
```diff
@@ -189,13 +224,23 @@ impl Weather {
     }
 
 
-    pub fn update(&mut self, camera: &Camera) {
+    pub fn update(&mut self, camera: &Camera, time: f32, world_manager: &super::WorldManager) {
+        if self.time == time {
+            return;
+        } else {
+            self.time += 1.0;
+        }
```
LukasKalbertodt: What is `time` doing here? Is this a frame-delta (time since the last frame)? If so, please call it delta. And `self.time += 1.0` seems wrong to me...

client/src/weather/mod.rs
```diff
@@ -223,6 +269,14 @@ impl Weather {
                 Form::Rain => {
                     instance.position[2] = instance.position[2] - ((particle.velocity + 0.5) * 1.0)
                 }
+                Form::Pollen => {
+                    instance.position[0] += particle.trans_x.sin() * 0.025;
+                    instance.position[1] += particle.trans_y.cos() * 0.025;
+                    instance.position[2] += (particle.trans_z.sin() - 0.7) * 0.01;
+                    particle.trans_y += 3.14 * 0.005 * (rand::random::<f32>() - 0.5);
+                    particle.trans_x += 3.14 * 0.005 * (rand::random::<f32>() - 0.5);
+                    particle.trans_z += 3.14 * 0.005 * (rand::random::<f32>() - 0.1);
```
LukasKalbertodt: Again, use `std::f32::consts::PI`

PR #121, opened 2016-07-29
Sun path in relation to day of the year

FlorianJanosch:


client/src/daytime.rs
```diff
@@ -82,12 +87,19 @@ impl DayTime {
...
         let half_day = DAY_LENGTH as f32 / 2.0;
 
         let theta;
         let phi;
+
+        let mut month_diff = self.time_day as f32 - half_year;//
+        if month_diff < 0.0 {
+            //
+            month_diff *= -1.0//
+        }//
```
LukasKalbertodt: Empty comments?

client/src/daytime.rs
```diff
@@ -32,7 +32,12 @@ impl Default for DayTime {
 const DAY_LENGTH: f32 = 720.0;
 const YEAR_LENGTH: u32 = 12;
 
-const MONTH_DIFFERENCE: f32 = 3.1415 / 18.0;
+
+// lengthens the day by a static offset
+const DAY_LENGTHER: f32 = 0.0;
```
LukasKalbertodt: Is this constant even used? Zero doesn't make sense to me...

client/src/daytime.rs
```diff
@@ -82,12 +87,19 @@ impl DayTime {
     /// returns the position of the sun corresponding to time
     /// only mid summer
     pub fn get_sun_position(&self) -> Vector3f {
-        // 0 degrees for mid summer
-        // 60 degrees for hard winter
+
+        let half_year = YEAR_LENGTH as f32 / 2.0; //
```
LukasKalbertodt: Empty comment?

client/src/daytime.rs
FlorianJanosch: This constant could be used for a static offset to lengthen the day later, but it could also be removed

PR #124, opened 2016-07-29
Basic world shadowing

jonas-schievink:
Plants are not yet affected by shadows (but they *do* cast shadows). We'll do that later, after we wrote a fragment shader we like.

### DIFF
--- a/client/shader/chunk_std.frag
+++ b/client/shader/chunk_std.frag
@@ -1,12 +1,25 @@
 #version 140
 
+in vec4 shadowCoord;
 in vec3 x_color;
 in vec3 surfaceNormal;
 
 out vec3 color;
 
+uniform sampler2D shadow_map;
+
 const vec3 sun = normalize(vec3(1.0, 0.0, 1.0));
+const float SHADOW_BIAS = 0.005;    // Prevent "acne" :D
+const float AMBIENT = 0.1;
+
 void main() {
     float diffuse = max(0.0, dot(sun, surfaceNormal));
-    color = x_color * 0.1 + x_color * diffuse;
+    color = x_color * AMBIENT + x_color * diffuse;
+
+    vec3 lightCoords = shadowCoord.xyz / shadowCoord.w;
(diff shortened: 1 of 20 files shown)


jonas-schievink:
Oh oops, I accidentally the whole branch :bomb: :boom:

jonas-schievink:
Yay for `reflog`!

client/shader/chunk_std.frag
```diff
@@ -1,12 +1,47 @@
...
+            shadowFactor += shadow ? 1.0 : 0.0;
+        }
+    }
+
+    // Height/Width of the square we sample for Percentage Closer Filtering
+    // (in Pixels)
+    const int PCF_PIXELS = 1 + 2 * SHADOW_PCF_RADIUS;
+
+    // Divide by number of pixels we sampled, to get  a range from 0 to 1
+    shadowFactor /= PCF_PIXELS * PCF_PIXELS;
```
LukasKalbertodt: `PCF_PIXELS` is a strange name when it holds the square root of the number of all PCF pixels... but yeah ok

LukasKalbertodt:
You're using a real depth framebuffer, which OpenGL writes`gl_Position.z/gl_Position.w` to. This means the depth distribution is logarithmic, which could decrease accuracy at the back plane. However, this might not matter at all...

client/shader/chunk_std.frag
```diff
@@ -1,12 +1,47 @@
...
 in vec3 x_color;
 in vec3 surfaceNormal;
 
 out vec3 color;
 
+// FIXME This should be a `sampler2DShadow`, but glium doesn't expose it
+uniform sampler2D shadow_map;
+// Shadow map height/width in pixels:
+// TODO Set from Rust code so it's always consistent
+uniform float shadow_map_size = 1024;
```
LukasKalbertodt: I think `textureSize(shadow_map, 0)` should return the correct resolution.

PR #128, opened 2016-07-29
Dynamic sky with sunrise and sunset

FlorianJanosch:


LukasKalbertodt:
This is not quite how I imagined it. Check if you can simplify your shader code still. GLSL functions like `lerp` could help. Shaders also don't really like many branches (if, else) -- it hurts the performance. I could later check how performance heavy your sky is.

And as far as I understood it, you won't implement *real*, simulated atmospheric scattering, right? That's fine, but please skim a few papers (for example the ones linked [here](http://vterrain.org/Atmosphere/)) to explain your decision not to implement any of those methods. This explanation would be very interesting in the final presentation.

jonas-schievink:
They tested performance - Didn't seem to have an impact (somewhat surprisingly)

FlorianJanosch:
`lerp` is called `mix` in [OpenGL](https://www.opengl.org/sdk/docs/man/html/mix.xhtml) (just for future reference)

PR #129, opened 2016-07-29
make use of dimension type for aspect ratio, fix #99

### LINKED ISSUE #99: Perspective matrix should use the correct aspect ratio
The aspect ratio of 16/9 to calculate the projection matrix is hardcoded in the `Camera` type right now. This needs to change!

vab9:


vab9:
should pass all tests now...
one of the tests was needed to use floating point division

client/src/camera.rs
```diff
@@ -17,13 +18,24 @@ impl Default for Camera {
             position: Point3::new(0.0, 0.0, 60.0),
             phi: -0.27,
             theta: 2.6,
+            aspect_ratio: 16.0 / 9.0,
```
jonas-schievink: I think it's better to remove the default impl instead. The concept of a default doesn't really make sense for a camera anyways.

vab9: will do

LukasKalbertodt:
Without having really looked through this PR, I think it still breaks when the window is resized. @Cranc implemented a framebuffer resize on window-resize, so we should also adjust the aspect ratio. This could be a bit harder, though.

PR #130, opened 2016-07-30
Use 3 codegen units.

jonas-schievink:
On my desktop PC (4+4 cores i7), this decreased the compile times after
changing `client` from 19.9 to 9.8 seconds (!).


jonas-schievink:
FPS is unaffected.

jonas-schievink:
Merging now or I'll lose my mind

PR #131, opened 2016-08-01
Outsource of the noise function

HelenaKeller:


base/src/lib.rs
```diff
@@ -4,9 +4,9 @@
 //!   defines a few own type
 //! - the world module is all about saving and managing the game world
 
-extern crate rand;
+pub extern crate rand;
 extern crate num_traits;
-extern crate noise;
+pub extern crate noise;
```
LukasKalbertodt: This shouldn't be necessary 0_o I didn't even know this is allowed

client/src/world/chunk_renderer.rs
```diff
@@ -36,6 +40,11 @@ impl ChunkRenderer {
                                           PrimitiveType::TrianglesList,
                                           &indices)
                 .unwrap(),
+            noise_map: match Texture2d::new(context.get_facade(),
+                                            tex_generator::create_noise(2 as u64)) {
+                Ok(p) => p,
+                Err(_) => panic!("did not work"),
+            },
```
LukasKalbertodt: This is what `try!` is for. You can use it instead of the `match`.

client/src/world/chunk_view.rs
```diff
@@ -43,6 +44,18 @@ impl ChunkView {
...
+        // let data = vec![
+        //     vec![(0u8, 0u8, 0u8), (0u8, 0u8, 255u8), (0u8, 0u8, 255u8)],
+        //     vec![(0u8, 0u8, 0u8), (0u8, 0u8, 255u8), (0u8, 0u8, 255u8)],
+        //     vec![(255u8, 0u8, 0u8), (0u8, 255u8, 0u8), (0u8, 255u8, 0u8)],
+        // ];
+
+        // let data = vec![
+        //         vec![(0u8, 0u8, 0u8)],
+        //         vec![(255u8, 255u8, 255u8)],
+        //     ];
```
LukasKalbertodt: This can be removed, right?

client/src/world/chunk_view.rs
```diff
@@ -51,9 +64,14 @@ impl ChunkView {
     }
 
     pub fn draw<S: glium::Surface>(&self, surface: &mut S, camera: &Camera) {
+
+
+
```
LukasKalbertodt: Random empty lines...

client/src/world/chunk_view.rs
```diff
@@ -51,9 +64,14 @@ impl ChunkView {
...
         let uniforms = uniform! {
             proj_matrix: camera.proj_matrix().to_arr(),
             view_matrix: camera.view_matrix().to_arr(),
+            my_texture:  self.renderer.noise_map.sampled().minify_filter(MinifySamplerFilter::NearestMipmapLinear)
+                .wrap_function(::glium::uniforms::SamplerWrapFunction::Repeat),
```
LukasKalbertodt: The `wrap_function` shouldn't be necessary, since you never fetch texels outside of the [0, 1] range. Ehm... but ok, it can stay.

base/src/lib.rs
jonas-schievink: It saves another `Cargo.toml` entry. Not sure what's better. `pub extern crate` was stabilized some time ago.

PR #133, opened 2016-08-01
plant_gen + biome at HexPillar

adRichter:


base/src/gen/plant/mod.rs
```diff
@@ -19,10 +20,14 @@ impl PlantGenerator {
...
+    pub fn rand<R: Rng>(rng: &mut R, biome: Biome) -> Self {
+        PlantGenerator::Tree(TreeGen::rand(rng, biome))
     }
 }
+
+// impl Rand for PlantGenerator {
+//     fn rand<R: Rng>(rng: &mut R) -> Self {
+//         PlantGenerator::Tree(TreeGen::rand(rng))
+//     }
+// }
```
jonas-schievink: This can be removed

client/src/renderer.rs
```diff
@@ -157,7 +157,7 @@ impl Renderer {
         // 0: Disable Bloom
         // 1: Enable Bloom
         // 2: Show only Bloom Map
-        let bloom_state = 1;
+        let bloom_state = 0;
```
jonas-schievink: Do you need to globally disable it? Why?

base/src/gen/plant/tree.rs
```diff
@@ -266,20 +423,33 @@ impl TreeGen {
...
+        };
+        let preset = rng.choose(matched_preset).unwrap().clone();
 
         TreeGen {
             preset: preset,
             branches: Vec::new(),
         }
+
+
+
```
LukasKalbertodt: Random empty lines?

client/src/renderer.rs
LukasKalbertodt: Yep, please enable it on master :stuck_out_tongue:

PR #135, opened 2016-08-01
prepare plant instancing

adRichter:


base/src/gen/world/mod.rs
```diff
@@ -175,18 +177,37 @@ impl ChunkProvider for WorldGenerator {
...
                     Some(section) => section.top,
                     None => HeightType::from_units(0),
                 };
 
                 props.push(Prop {
                     baseline: height,
-                    prop: PropType::Plant(gen.generate(&mut rng)),
+                    prop: PropType::Plant(PlantGenerator::new(PlantType::OakTree)
+                        .generate(&mut rng)),
+                    plant_index: plant_index as usize,
```
jonas-schievink: Why not put a `PlantType` in there? That way it's more type safe.

PR #137, opened 2016-08-01
fixed the fragmentshader in chunk_std.frag

RomanSchott:


client/shader/chunk_std.vert
```diff
@@ -26,7 +26,7 @@ void main() {
         position.z * height + offset.z,
         1);
     gl_Position = proj_matrix * view_matrix * world_coords;
-    shadowCoord = depth_view_proj * world_coords;
+    shadowCoordT = depth_view_proj * world_coords;
```
LukasKalbertodt: Why was it necessary to rename this? 0.o

LukasKalbertodt:
Already fixed in master

Issue #136, opened 2016-08-01
view frustum culling to improve performance #RenderWhatMatters


Cranc:
I'm currently working on a basic frustum implementation which later on can be used to improve performance of rendering, by not rendering the entire scene but only the stuff inside the players view.

- [ ] create basic frustum class.
- [ ] impelement the frustum inside renderer.
- [ ] pray that it increases performance a lot.

current stage: [here](https://gist.github.com/Cranc/f55fa1169fc55f71bbb2f8484c72b6ff )

Cranc:
needs still some fine tuning and currently does not seem to improve performance much.

Cranc:
does anyone have a clue why my view frustum culling seems to not register the chunks to render correctly.

Also i  could be paranoid but i think it also renders behind me xD no way to proof that though

[image](https://www.dropbox.com/s/8fgp7iuta148ws2/Screenshot%202016-08-03%2011.49.52.png?dl=0)

PR #147, opened 2016-08-02
implementation of frustum rendering

Cranc:
part of #136


Cranc:
```
=== Checking Rust style with rustfmt... ==============
! incorrectly formatted: client/src/world/chunk_view.rs
! incorrectly formatted: client/src/world/world_view.rs
! incorrectly formatted: client/src/renderer.rs
```

did someone changed rustfmt recently because since i work on win10 i do most myself
and check with https://play.rust-lang.org/ before commit and nothing on my end indicated something.

LukasKalbertodt:
Nope, it should still be `rustfmt v0.5.0`...

Cranc:
for now i will simply ignore this rustfmt stuff cause i can't tell whats exactly wrong ,,, going back to optimizing


small note: i went and made sure the game runs on my gtx970M instead of intel (60+ fps 800*600 and 45fps fullscreen 1920*1080)

Cranc:
I added a SimpleCull to furstrum.rs which simply is a "fov culling" but this at least works fine also implemented it already so it is on right now.

Cranc:
if someone can tell me where in chunk_view.rs the format error is that would be nice =)

`! incorrectly formatted: client/src/world/chunk_view.rs`

Cranc:
closed cause #168 solves format error

PR #150, opened 2016-08-02
add stars with noise map

FlorianJanosch:


client/src/view/sky_view.rs
```diff
@@ -45,11 +52,32 @@ impl SkyView {
...
+
         SkyView {
             vertex_buffer: vbuf,
             index_buffer: ibuf,
             program: context.load_program("skydome").unwrap(),
             sun_position: Point3f::new(0.0, 0.0, -1000.0),
+            star_map: match Texture2d::new(context.get_facade(), v) {
+                Ok(p) => p,
+                Err(_) => panic!("Could not load stars"),
+            },
```
LukasKalbertodt: That's what `expect()` is for. https://doc.rust-lang.org/core/result/enum.Result.html#method.expect

client/src/view/sky_view.rs
```diff
@@ -45,11 +52,32 @@ impl SkyView {
                                     &raw_index_buffer)
             .unwrap();
 
+        let seed = 54342354434;
+        let mut star_rng = seeded_rng(seed, 0, ());
+        let star_table = PermutationTable::rand(&mut star_rng);
+
+
+        // values between 0 and 0.5
+        const TEX_SIZE: usize = 3000;
```
LukasKalbertodt: Ehm... I think thats a bit large. Also: use 2^k texture sizes please. Maybe 2048

FlorianJanosch:
@jonas-schievink never do `git add .` :rage: :hankey: :neckbeard:

PR #176, opened 2016-08-04
removed debug println

mmildt:


jonas-schievink:
This is already removed on master

PR #185, opened 2016-08-04
Fixes syntax errors in shaders

jovobe:
Now it runs on OS X too 


jonas-schievink:
We should really put a GLSL validator in the CI scripts...

jonas-schievink:
Opened #187

Issue #88, opened 2016-07-25
Dynamic Chunk loading is off by a bit


jonas-schievink:
You can see it quite easily: Just fly above the map and look down while moving. In some directions, chunks are loaded closer to the player, while in others they're loaded farther away.

The chunk positions are just missing an offset so that the position we use for the calculation is in the middle of them instead of being on a corner. Easy fix, just needs a good eye to see if it's fixed. Also a good way to get into how dynamic chunk loading works if anyone's interested.

jonas-schievink:
Not easy, this is a more intricate issue (it works well when in positive chunk coordinates but breaks when entering negative coordinates). Might be a bug in `from_real`.

PR #190, opened 2016-08-07
One vertex buffer to rule them all

LukasKalbertodt:
Instead of using instancing, this PR saves a whole chunk in one vertex and index buffer. This increases the performance of rendering the ground a lot. I lack exact benchmarks, but I hope a few of you could provide such benchmarks. 

Without rendering shadow or plants, I get between 50 and 60 FPS on my integrated Intel HD Graphics 530. With plants enabled I still get around 50 FPS in biomes with few plants, but FPS drops a lot in forests.

There is still a lot of room for improvements, but I figured this is enough for now. 

Oh and by the way: I fixed the indexing bug. Or at least *one* big indexing bug.

### DIFF
--- a/base/src/math/axial_point.rs
+++ b/base/src/math/axial_point.rs
@@ -31,34 +31,37 @@ impl AxialPoint {
     /// system using `world::{HEX_INNER_RADIUS, HEX_OUTER_RADIUS}`.
     pub fn to_real(&self) -> Point2f {
         Point2f {
-            x: ((2 * self.q + self.r) as DefaultFloat) * HEX_INNER_RADIUS,
-            y: -(self.r as DefaultFloat) * (3.0 / 2.0) * HEX_OUTER_RADIUS,
+            x: ((2 * self.q - self.r) as DefaultFloat) * HEX_INNER_RADIUS,
+            y: (self.r as DefaultFloat) * (3.0 / 2.0) * HEX_OUTER_RADIUS,
         }
     }
 
     /// Return the `AxialPoint` from a `Point2f`
     pub fn from_real(real: Point2f) -> Self {
-        let q: f32 = (real.x * ::math::SQRT_3 / 3.0 - real.y / 3.0) / HEX_OUTER_RADIUS;
-        let r: f32 = (real.y * 2.0 / 3.0) / HEX_OUTER_RADIUS;
+        let q = (real.x * ::math::SQRT_3 / 3.0 - real.y / 3.0) / HEX_OUTER_RADIUS;
+        let r = (real.y * 2.0 / 3.0) / HEX_OUTER_RADIUS;
 
-        let y: f32 = -q - r;
+        let s = -q - r;
 
         // Rounding
-        let mut rx: i32 = (q + 0.5) as i32;
-        let ry: i32 = (y + 0.5) as i32;
-        let mut rz: i32 = (r + 0.5) as i32;
+        let mut rq = q.round();
+        let mut rr = r.round();
+        let mut rs = s.round();
 
         // To test the right rounding
-        let x_diff = (rx as f32 - q).abs();
-        let y_diff = (ry as f32 - y).abs();
-        let z_diff = (rz as f32 - r).abs();
-        if x_diff > y_diff && x_diff > z_diff {
-            rx = -ry - rz;
-        } else if y_diff <= z_diff {
-            rz = -rx - ry;
+        let q_diff = (rq - q).abs();
+        let r_diff = (rr - r).abs();
+        let s_diff = (rs - s).abs();
+
+        if q_diff > r_diff && q_diff > s_diff {
+            rq = -rr - rs;
+        } else if s_diff > r_diff {
+            rr = -rq - rs;
+        } else {
+            rs = -rq - rr;
         }
 
-        AxialPoint { q: rx, r: rz }
+        AxialPoint { q: -rs as AxialType, r: rr as AxialType}
     }
 
     /// Returns the `s` component of corresponding cube coordinates. In cube
--- a/base/src/math/axial_point.rs
+++ b/base/src/math/axial_point.rs
@@ -44,7 +44,7 @@ impl AxialPoint {
         let s = -q - r;
 
         // Rounding
-        let mut rq = q.round();
+        let rq = q.round();
         let mut rr = r.round();
         let mut rs = s.round();
 
@@ -54,7 +54,7 @@ impl AxialPoint {
         let s_diff = (rs - s).abs();
 
         if q_diff > r_diff && q_diff > s_diff {
-            rq = -rr - rs;
+            // rq = -rr - rs;
         } else if s_diff > r_diff {
             rr = -rq - rs;
         } else {
--- a/base/src/math/axial_point.rs
+++ b/base/src/math/axial_point.rs
@@ -32,7 +32,7 @@ impl AxialPoint {
     pub fn to_real(&self) -> Point2f {
         Point2f {
             x: ((2 * self.q + self.r) as DefaultFloat) * HEX_INNER_RADIUS,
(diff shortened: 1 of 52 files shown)


base/src/math/axial_point.rs
```diff
@@ -31,34 +31,37 @@ impl AxialPoint {
     /// system using `world::{HEX_INNER_RADIUS, HEX_OUTER_RADIUS}`.
     pub fn to_real(&self) -> Point2f {
         Point2f {
-            x: ((2 * self.q + self.r) as DefaultFloat) * HEX_INNER_RADIUS,
+            x: ((2 * self.q - self.r) as DefaultFloat) * HEX_INNER_RADIUS,
```
jonas-schievink: wow, not even this was correct?

client/shader/chunk_std.vert
```diff
@@ -32,12 +26,13 @@ out vec3 pos;
 uniform mat4 proj_matrix;
 uniform mat4 view_matrix;
 uniform mat4 depth_view_proj;
+uniform vec2 offset;
```
jonas-schievink: This seems unused?

client/shader/chunk_shadow.vert
```diff
@@ -4,20 +4,14 @@
...
-// Per-instance attributes:
-// Height in units, not world coordinates, since the "pillar prototype" has a
-// height of one unit.
-in float height;
-in vec3 offset;
-//in vec3 material_color;
-
 uniform mat4 proj_matrix;
 uniform mat4 view_matrix;
+uniform vec2 offset;
```
jonas-schievink: This looks unused too

LukasKalbertodt: Uhm... what about ...

client/shader/chunk_shadow.vert
```diff
@@ -4,20 +4,14 @@
...
 
 void main() {
     vec4 world_coords = vec4(
         position.xy + offset.xy,
```
LukasKalbertodt: ... here?

client/src/world/chunk_view.rs
```diff
@@ -92,50 +70,28 @@ impl ChunkView {
                 test: DepthTest::IfLess,
                 ..Default::default()
             },
-            backface_culling: BackfaceCullingMode::CullClockwise,
+            backface_culling: BackfaceCullingMode::CullCounterClockwise,
```
jonas-schievink: This was intentional and is supposed to reduce shadowing artifacts. Not sure how much it matters with variance shadow mapping, though.

client/shader/chunk_std.vert
```diff
@@ -48,15 +43,10 @@ void main() {
...
-    if(tex_coords.y > 1.5){
-        x_tex_coords.y = height;
-    }
-
     x_position = position;
     x_ground = ground;
 
     // position in relation to player ignoring hight, needed for fog
-    pos = vec3(offset - cam_pos);
+    // pos = vec3(offset - cam_pos);
```
jonas-schievink: This makes the line below read `pos` before it is initialized, which causes the rendering glitches I was experiencing.

client/shader/chunk_shadow.vert
jonas-schievink: Oh, nevermind :laughing:

base/src/math/axial_point.rs
LukasKalbertodt: Well... there are many "correct" versions. In the version of the axial coordinate system that I chose this was wrong, yes.

jonas-schievink:
The indexing fixes sadly do not fix #88

base/src/math/axial_point.rs
```diff
@@ -31,34 +31,37 @@ impl AxialPoint {
...
-        if x_diff > y_diff && x_diff > z_diff {
-            rx = -ry - rz;
-        } else if y_diff <= z_diff {
-            rz = -rx - ry;
+        let q_diff = (rq - q).abs();
+        let r_diff = (rr - r).abs();
+        let s_diff = (rs - s).abs();
+
+        if q_diff > r_diff && q_diff > s_diff {
+            // rq = -rr - rs;
```
jonas-schievink: Should this stay or should this go?

client/src/world/world_view.rs
```diff
@@ -154,9 +153,9 @@ impl WorldView {
             };
 
             let outline_uniforms = uniform! {
-              outline_pos: self.outline.position().to_arr(),
-              proj_matrix: camera.proj_matrix().to_arr(),
-              view_matrix: camera.view_matrix().to_arr()
+                outline_pos: self.outline.position().to_arr(),
+                proj_matrix: camera.proj_matrix().to_arr(),
+                view_matrix: camera.view_matrix().to_arr()
```
jonas-schievink: I like how rustfmt didn't say anything just because it's in a macro...

client/src/world/chunk_view.rs
```diff
@@ -240,20 +177,626 @@ pub struct Vertex {
...
+    let (ca, cb) = EDGE_CORNERS_TO_NEIGHBOR[dir.idx()];
+    let normal = EDGE_NORMALS[dir.idx()];
+    let corner_cw = [
+        (offset.to_real() + ca * HEX_OUTER_RADIUS, 0.25),
+        (offset.to_real() + cb * HEX_OUTER_RADIUS, 0.75),
+    ];
+
+    // In order to have correctly stretched textures, we have to change one
+    // texture coordinate depending on the height.
+    let v_bottom = 0.5 * (top.to_real() - bottom.to_real()) / PILLAR_STEP_HEIGHT;
```
jonas-schievink: Is that 0.5 just meant to stretch the texture an additional bit so it looks better on our small pillar steps?

client/src/world/chunk_view.rs
```diff
@@ -240,20 +177,626 @@ pub struct Vertex {
...
+/// This function adds a side with the given parameter.
+fn add_side(
+    bottom: HeightType,
+    top: HeightType,
+    ground: GroundMaterial,
+    normal_to_a: bool,
+    dir: EdgeDir,
+    offset: AxialPoint,
+    vertices: &mut Vec<Vertex>,
+    indices: &mut Vec<u32>,
```
jonas-schievink: Wouldn't hurt to document the parameters in the doc comment

jonas-schievink: Especially `normal_to_a` isn't obvious

jonas-schievink:
The buffer building code is a tad too complex for my liking, but very well documented so looks good to me

client/src/world/chunk_view.rs
LukasKalbertodt: If `PILLAR_STEP_HEIGHT` would be equal to `HEX_OUTER_RADIUS` (assume both are 1), then a side piece with the height `PILLAR_STEP_HEIGHT` is quadratic and should have `max(u) - min(u) == max(v) - min(v)`. But since the `HEX_OUTER_RADIUS` is a *radius* not diameter, it's only half of the whole 0-1 texture.

Does this make sense? :stuck_out_tongue:

base/src/math/axial_point.rs
LukasKalbertodt: I think I actually want this to stay, since it then represents the algorithm from redblobgames.com more closely. But I can add a comment. Will do it tomorrow when fixing the other stuff.

LukasKalbertodt:
> The buffer building code is a tad too complex for my liking, but very well documented so looks good to me.

I think so, too, but I wasn't able to come up with a less complex solution yesterday that does exactly the same thing. But feel free to simplify it :stuck_out_tongue: I'd like to tweak it a lot still, but I have other things to do.

LukasKalbertodt:
Btw, I just checked: this PR introduces the flickering plants. That's pretty *nice*, since the PR doesn't change any of the plant rendering code (except for the two deleted empty lines). OpenGL, fuck yeah :flushed: 

We should check on what GPUs the plant flickering occurs.

PR #191, opened 2016-08-07
Allow disabling tessellation and simplify tcs

jonas-schievink:
I just copied most of the shader together so it works. Gonna clean up all the shader handling soon.


vab9:
Awesome! I tried to do this too but failed miserably...

vab9:
Should we merge this soon?

jonas-schievink:
I'm sure @LukasKalbertodt doesn't mind

Issue #187, opened 2016-08-05
CI: Validate Shaders


jonas-schievink:
We've had many issues where the nVidia driver would accept our shaders while other drivers (mainly Intel) would reject it (rightfully so). We should use the reference compiler (https://github.com/KhronosGroup/glslang) to validate our shaders with Travis instead.

Unfortunately I haven't yet found an easy way to install the compiler in our Travis environment (without using `apt-get`), so we might lose the fast builds of the container-based infrastructure.

PR #194, opened 2016-08-09
Validate shaders in Travis. Hope this will fix issue #187.

vab9:
Not 100% if this will work. The shaders throw erros with glslangValidator on osx as it is, but that might mean nothing for Travis.

I'm also not sure if caching the validator is a good idea or not...

### DIFF
--- a/ci/install-glslangValidator.sh
+++ b/ci/install-glslangValidator.sh
@@ -1,0 +1,15 @@
+#!/bin/bash
+set -e
+
+# check for validator folder (travis caching)
+if [ ! -d "validator/" ]; then
+    echo "need to install glslangValidator..."
+    mkdir validator/
+    cd validator/
+    wget "https://cvs.khronos.org/svn/repos/ogl/trunk/ecosystem/public/sdk/tools/glslang/Install/Linux/glslangValidator"
+    chmod +x glslangValidator
+    cd ../
+    echo "installed glslangValidator to $PWD/validator/glslangValidator"
+else
+    echo "using cached glslangValidator"
+fi
--- a/ci/install-glslangValidator.sh
+++ b/ci/install-glslangValidator.sh
@@ -1,15 +1,23 @@
 #!/bin/bash
 set -e
 
+echo "need to install glslangValidator..."
+mkdir validator/
+cd validator/
+wget "https://cvs.khronos.org/svn/repos/ogl/trunk/ecosystem/public/sdk/tools/glslang/Install/Linux/glslangValidator"
+chmod +x glslangValidator
+cd ../
+echo "installed glslangValidator to $PWD/validator/glslangValidator"
+
 # check for validator folder (travis caching)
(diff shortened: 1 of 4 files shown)


ci/install-glslangValidator.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+set -e
+
+echo "need to install glslangValidator..."
+mkdir validator/
+cd validator/
+wget "https://cvs.khronos.org/svn/repos/ogl/trunk/ecosystem/public/sdk/tools/glslang/Install/Linux/glslangValidator"
```
jonas-schievink: Oh that's neat! I feared we'd have to compile it on our own.

ci/validate-shaders.sh
```diff
@@ -0,0 +1,42 @@
+#!/bin/bash
+
+set -o errexit -o nounset
+
+echo ""
+echo "=== Validating Shaders =============="
+
+SHADER_DIR=`find $PWD -name shader`
```
jonas-schievink: I'd prefer to scan for shader files in the whole tree, like `check-basic-style.sh` does. You can even use the same pattern: `FILES='.+\.\(rs\|vert\|tesc\|tese\|geom\|frag\|comp\)'` (Note: our tessellation shaders have a wrong suffix for some reason)

ci/validate-shaders.sh
```diff
@@ -0,0 +1,42 @@
...
+trap "rm $ERROR_TMP; exit" SIGHUP SIGINT SIGTERM
+
+# Make sure only shaders are in client/shader/ for this to work
+for shader in $SHADER_DIR/*; do
+    echo "Validating: `basename $shader` ..."
+    if ! glslangValidator -s $shader > /dev/null;
+        then
+        echo "=== ERRORS in `basename $shader` =============" >> $ERROR_TMP
+        glslangValidator $shader | grep "ERROR" >> $ERROR_TMP
+        echo "" >> $ERROR_TMP
```
jonas-schievink: Why the indirection through the temporary file? Theoretically you can just let `glslangValidator` output to stdout/stderr as normal and set an error flag if it fails. Take a look at the `check-basic-style.sh` script, it works similarly.

jonas-schievink:
Travis is apparently running Ubuntu 12.04, not sure how easy it will be to get the validator to run.

12.04.......

jonas-schievink:
By the way, all the `#N` comments are referencing random unrelated issues and PRs

jonas-schievink:
The glslang project itself seems to explicitly enable Ubuntu 14.04 LTS: https://github.com/KhronosGroup/glslang/blob/master/.travis.yml

Sadly, this would increase CI startup times by quite a bit (as we'll no longer be using the fast container-based infrastructure).

vab9:
Damn, ok! It was worth a try... Travis-CI is pretty cool though

PR #203, opened 2018-11-12
Update dependencies and stuff like that

### LINKED ISSUE #202: glium dependency has a broken link
Repository https://github.com/jonas-schievink/glium.git doesn't exist, but `client` has it as a dependency.

LukasKalbertodt:
This updates dependencies and fixes a few dependency links. See commit messages for more information. It now compiles with the newest stable (1.30) again. We get a ton of [these warnings](https://github.com/rust-lang/rust/issues/41620), but it works.

CC @jonas-schievink in case you want to take a look. Otherwise I'd just merge.

Closes #202


base/src/math/axial_vector.rs
```diff
@@ -12,7 +12,6 @@ use std::cmp;
 ///
 /// [hex-blog]: http://www.redblobgames.com/grids/hexagons/#coordinates
 #[derive(Clone, Copy, PartialEq, Eq, Hash)]
-#[repr(C, packed)]
```
jonas-schievink: This probably never did anything anyways

